PR #32, opened 2015-05-29
Cleaned up/Added tests

saghmrossi:


alabid:
I'd prefer to review one commit per pull request. All of these commits can be squashed into one using:
`git rebase -i <name of your upstream 1.0 branch>`

PR #36, opened 2015-05-29
Fix linking and spacing in travis config

kyeah:


alabid:
LGTM provided there are no errors. Next step is to fix the warnings. I'll merge once I see a green build.

PR #39, opened 2015-06-05
ConnectionString Implementation

kyeah:
Initial implementation for the Connection String URI parser. User-specified options are not validated, currently.

### DIFF
--- a/src/client/connstring.rs
+++ b/src/client/connstring.rs
@@ -2,7 +2,7 @@ use std::ascii::AsciiExt;
 use std::collections::HashMap;
 use std::vec::Vec;
 
-pub const DEFAULT_PORT: u32 = 27017;
+pub const DEFAULT_PORT: u16 = 27017;
 pub const URI_SCHEME: &'static str = "mongodb://";
 
 /// MongoDB connection types.
@@ -15,15 +15,15 @@ pub enum ConnectionType {
 
 /// Encapsulates the hostname and port of a host.
 pub struct Host {
-    pub hostname: String,
-    pub port: u32,
+    pub host_name: String,
+    pub port: u16,
 }
 
 impl Host {
     /// Creates a new Host struct.
-    pub fn new(hostname: String, port: u32) -> Host {
+    pub fn new(host_name: String, port: u16) -> Host {
         Host {
-            hostname: hostname,
+            host_name: host_name,
             port: port,
         }
     }
@@ -51,7 +51,7 @@ impl ConnectionOptions {
 
     /// Retrieves an option using a borrowed str
     pub fn get_str(&self, key: &str) -> Option<&str> {
-        match self.options.get(&key.to_string()) {
+        match self.options.get(&key.to_owned()) {
             Some(val) => Some(&val),
             None => None,
         }
@@ -72,8 +72,8 @@ pub struct ConnectionString {
 
 impl ConnectionString {
     /// Creates a new ConnectionString for a single, unreplicated host.
-    pub fn new(hostname: &str, port: u32) -> ConnectionString {
-        let host = Host::new(hostname.to_string(), port);
+    pub fn new(host_name: &str, port: u16) -> ConnectionString {
+        let host = Host::new(host_name.to_owned(), port);
         ConnectionString::new_from_host(host)
     }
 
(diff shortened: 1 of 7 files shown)


src/client/connstring.rs
```diff
@@ -0,0 +1,318 @@
...
+        match self.options.get(&key.to_owned()) {
+            Some(val) => Some(&val),
+            None => None,
+        }
+    }
+}
+
+/// Encapsulates information for connection to a single MongoDB host or replicated set.
+pub struct ConnectionString {
+    ctype: ConnectionType,
```
vkarpov15: I'm not sure of the purpose of this property, doesn't seem to really provide any useful information as far as I see. What was your reasoning for including this property?

src/client/connstring.rs
```diff
@@ -0,0 +1,318 @@
...
+    let mut password: Option<String> = None;
+    let mut database: Option<String> = Some("test".to_owned());
+    let mut collection: Option<String> = None;
+    let mut options: Option<ConnectionOptions> = None;
+
+    // Split on host/path
+    let (host_str, path_str) = match addr.contains(".sock") {
+        true => {
+            let (host_part, path_part) = rpartition(addr, "/");
+            let host_test_uri = &format!("{}{}", URI_SCHEME, host_part);
```
vkarpov15: Don't really like the extra string creation here, we should refactor this so that there's an internal function that parses connection string minus scheme if anything

src/client/connstring.rs
```diff
@@ -0,0 +1,318 @@
...
+            match parse(host_test_uri) {
```
vkarpov15: Not really a fan of this implementation - looks like you rely on parse() to fail if there's no database specified because the last slash will be before '.sock' seems a little brittle. May be better to rework this to split by '?' first, then by ',', and then split on '.sock' sorta like how muri does it: https://github.com/aheckmann/muri/blob/master/lib/index.js#L201

kyeah: +1. Muri also stores the ipcs in a separate ```ipc``` member instead of as ```host_name```, which is probably a better way to organize it.

src/client/connstring.rs
kyeah: Artifact of initially modeling it off the CXX driver; will remove, as it serves mostly to indicate # of hosts being connected to. Anything that needs this information can check hosts.len() instead.

vkarpov15: I'd recommend just removing it, we don't really use it internally, and it doesn't map well to how we want to think about connection strings going forward.

src/client/connstring.rs
```diff
@@ -0,0 +1,318 @@
...
+                Some(idx) => {
+                    let port = &entity[idx+2..];
+                    match port.parse::<u16>() {
+                        Ok(val) => Ok(Host::new(entity[1..idx].to_ascii_lowercase(), val)),
+                        Err(_) => Err("Port must be an integer."),
+                    }
+                },
+                None => Ok(Host::new(entity[1..].to_ascii_lowercase(), DEFAULT_PORT)),
+            }
+        }
```
alabid: Should there be a comma after this closing brace?

kyeah: Yep; the compiler didn't catch or complain about it, but I added it for consistency.

src/client/connstring.rs
kyeah: Removed with the refactored IPC socket parsing.

src/client/connstring.rs
jchinlee: +1: not necessary to have the trailing comma but easier to move cases around (even if there are only two in this situation, generally better stylistically).

vkarpov15:
lgtm on my end. @alabid ?

vkarpov15:
Hey @steveklabnik, still interested in helping us review PRs for general rustyness?

This PR is an implementation of our (currently internal) spec for parsing connection strings, like "mongodb://localhost:27017" so we can determine which server to connect to.

alabid:
LGTM. Good stuff.

steveklabnik:
@vkarpov15 yes, I am :) I will check the diff out.

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
+use std::ascii::AsciiExt;
+use std::collections::HashMap;
+use std::vec::Vec;
```
steveklabnik: This is [in the prelude](http://doc.rust-lang.org/stable/std/prelude/), and so does not need to be explicitly imported.

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
...
+    // Creates a new Host struct.
+    fn new(host_name: String, port: u16) -> Host {
+        Host {
+            host_name: host_name,
+            port: port,
+            ipc: String::new(),
+        }
+    }
+
+    fn new_ipc(ipc: String) -> Host {
```
steveklabnik: I would think about renaming this function. This makes it sound like it's making a new ipc, when it's making  a new `Host`. There's a small convention for alternate constructors like this: `with_*`. So I'd call it `with_ipc()`.

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// Creates a new ConnectionOptions struct.
+    pub fn new(options: HashMap<String, String>, read_pref_tags: Vec<String>) -> ConnectionOptions {
+        ConnectionOptions {
+            options: options,
+            read_pref_tags: read_pref_tags,
+        }
+    }
+
+    /// Retrieves an option using a borrowed String
+    pub fn get(&self, key: &String) -> Option<&String> {
```
steveklabnik: `&String` is almost never what you want. It's a pointer to a pointer, and you don't need the second indirection. Both parts of this signature should probably be `&str`.

`HashMap` specifically supports looking up `String` keys with `&str`: http://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#method.get

    fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
        where K: Borrow<Q>,
                  Q: Hash + Eq

This type signature is _intense_, but the important part is `K: Borrow<Q>`, in other words, the key's type supports `Borrow` for the argument's type. And `impl Borrow<str> for String` is in the standard library. So it works: http://is.gd/ywiEUu

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.options.get(key)
+    }
+
+    /// Retrieves an option using a borrowed str
+    pub fn get_str(&self, key: &str) -> Option<&str> {
+        match self.options.get(&key.to_owned()) {
```
steveklabnik: for similar reasons, you don't need the `to_owned()` here, I'd bet.

steveklabnik: in fact, this whole method is redundant if you modify `get`'s type signature :)

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+impl ConnectionString {
+    /// Creates a new ConnectionString for a single, unreplicated host.
+    pub fn new(host_name: &str, port: u16) -> ConnectionString {
+        let host = Host::new(host_name.to_owned(), port);
+        ConnectionString::new_from_host(host)
+    }
+
+    fn new_from_host(host: Host) -> ConnectionString {
```
steveklabnik: same here as above, i'd call this `with_host()`

src/client/connstring.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    let mut hosts: Vec<Host>;
+    let mut user: Option<String> = None;
+    let mut password: Option<String> = None;
+    let mut database: Option<String> = Some("test".to_owned());
+    let mut collection: Option<String> = None;
+    let mut options: Option<ConnectionOptions> = None;
+
+    // Split on host/path
+    let (host_str, path_str) = match addr.contains(".sock") {
```
steveklabnik: a `match` that returns `true` or `false` is probably better written as an `if`

tests/client/connstring.rs
```diff
@@ -0,0 +1,247 @@
+use mongodb::client::connstring;
+
+#[test]
+fn test_valid_uri() {
```
steveklabnik: including `test_` at the start of test functions isn't considered good style: you already have the attribute showing it's a test, and the runner uses the name to print it out, so you get output like

    test test_valid_uri

intead of

    test valid_uri

tests/client/connstring.rs
```diff
@@ -0,0 +1,247 @@
...
+        "mongodb://localhost:27017",
+        "mongodb://localhost:27017/",
+        "mongodb://localhost:27017/?",
+        "mongodb://127.0.0.1",
+        "mongodb://127.0.0.1/",
+        "mongodb://127.0.0.1/?",
+        "mongodb://127.0.0.1:27017",
+        "mongodb://127.0.0.1:27017/",
+        "mongodb://127.0.0.1:27017/?",
+        );
```
steveklabnik: I personally would have this `);` outdented to match up with the `let`, but that's subjective.

tests/client/connstring.rs
```diff
@@ -0,0 +1,247 @@
...
+
+#[test]
+fn test_invalid_prefix() {
+    let invalid_uris = vec!(
+        "mongodb:/localhost",
+        "mngodb://localhost",
+        "mongodb//localhost",
+        "://localhost",
+        "localhost:27017",
+        );
```
steveklabnik: same here

steveklabnik:
Comments left! Overall, this code seems great, just a few things :)

kyeah:
Thanks for the feedback, @steveklabnik! I'll have a new PR out to get these changes in.

PR #40, opened 2015-06-08
Rust style conventions for connstring

kyeah:
Stylistic changes recommended by Steve in #39.

### DIFF
--- a/src/client/connstring.rs
+++ b/src/client/connstring.rs
@@ -1,6 +1,5 @@
 use std::ascii::AsciiExt;
 use std::collections::HashMap;
-use std::vec::Vec;
 
 pub const DEFAULT_PORT: u16 = 27017;
 pub const URI_SCHEME: &'static str = "mongodb://";
@@ -22,7 +21,7 @@ impl Host {
         }
     }
 
-    fn new_ipc(ipc: String) -> Host {
+    fn with_ipc(ipc: String) -> Host {
         Host {
             host_name: String::new(),
             port: DEFAULT_PORT,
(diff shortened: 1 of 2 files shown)


src/client/connstring.rs
```diff
@@ -22,7 +21,7 @@ impl Host {
         }
     }
 
-    fn new_ipc(ipc: String) -> Host {
+    fn with_ipc(ipc: String) -> Host {
```
vkarpov15: Minor nitpick: I think what steve meant was `new_with_ipc` is more idiomatic rust.

src/client/connstring.rs
```diff
@@ -79,10 +70,10 @@ impl ConnectionString {
     /// Creates a new ConnectionString for a single, unreplicated host.
     pub fn new(host_name: &str, port: u16) -> ConnectionString {
         let host = Host::new(host_name.to_owned(), port);
-        ConnectionString::new_from_host(host)
+        ConnectionString::with_host(host)
     }
 
-    fn new_from_host(host: Host) -> ConnectionString {
+    fn with_host(host: Host) -> ConnectionString {
```
vkarpov15: Ditto, think this should be `new_with_host`

src/client/connstring.rs
kyeah: ack - fixing.

kyeah: Actually, should be ```with_*```. std::vec::Vec provides a [```with_capacity``` constructor](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity).

PR #41, opened 2015-06-08
added wire protocol module and documentation

saghmrossi:


src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,309 @@
...
+/// Gets the length of a BSON document.
+//
+/// # Arguments
+///
+/// `bson` - The BSON document to find the length of.
+///
+/// # Return value
+///
+/// Returns the number of bytes in the serialized BSON document.
+fn length_of_bson(bson: &BSON) -> Result<i32, String> {
```
vkarpov15: I'm pretty sure it's more rusty to implement this as a trait on top of the bson doc

tests/client/wire_protocol.rs
```diff
@@ -0,0 +1,40 @@
...
+                Err(_) => panic!("Could not create message!")
+            };
+
+            match cm.write(&mut stream) {
+                Ok(_) => (),
+                Err(s) => panic!("{}", s)
+            };
+
+            match Message::read(&mut stream) {
+                Ok(bson) => match bson.get("foo") {
```
vkarpov15: nested matches are not great for readability + maintainability, I'd put this behind a separate variable

Cargo.toml
```diff
@@ -1,8 +1,10 @@
...
 version = "0.1.0"
-authors = ["Valeri Karpov <valkar207@gmail.com>"]
+authors = ["Valeri Karpov <valkar207@gmail.com>",
+           "Saghm Rossi <saghmrossi@gmail.com",
+           "Kevin Yeh <kevinyeah@utexas.edu"]
 
 [dependencies]
-
-libc="0.1.8"
+byteorder = "*"
```
vkarpov15: Not sure about rust, but using my node-based knee-jerk reaction is 'dont use `*` ever'. In rust '~0.3' and '~0.1' should be sufficient because of cargo.lock. Also, small nitpick, but it helps to keep deps in alphabetical order.

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///                     returned by the query.
+    /// `return_field_selector - An optional projection of which fields should
+    ///                          be present in the documents to be returned by
+    ///                          the query.
+    ///
+    /// # Return value
+    ///
+    /// Returns nothing on success, or an error string on failure.
+    fn write_query(buffer: &mut Write, header: &Header,
+                   flags: i32, full_collection_name: &str,
```
christkv: it might make for easier comprehension if flags are broken out into a boolean struct instead of just a i32 value delegating the encoding to a higher level of the driver.

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// `return_field_selector - An optional projection of which fields should be
+    ///                          present in the documents to be returned by the
+    ///                          query.
+    ///
+    /// # Return value
+    ///
+    /// Returns the newly-created Message.
+    pub fn new_query(header_request_id: i32, flags: i32,
+                     full_collection_name: String, number_to_skip: i32,
+                     number_to_return: i32, query: BSON,
```
jedestep: This method should not take ownership of the `query` parameter, unless it is really the case that the lifetime of the query should end here (which to me it seems like it shouldn't).

src/client/wire_protocol/operations.rs
saghmrossi: Yeah, I thought about that, and I probably should have done it. I was in a rush when I added this since it was a fix to the bug where the server was expecting a message that was too long.

Cargo.toml
saghmrossi: Ah, okay, sorry about that. I was just copying the dependency line that they suggested on github.

src/client/wire_protocol/operations.rs
saghmrossi: If I changed it to borrow `query`, would I still be able to have the struct that's returned take ownership of it? I thought that since the field in the struct isn't a pointer then I should have the method take ownership of it.

src/client/wire_protocol/header.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// `buffer` - The buffer to write to.
+    ///
+    /// # Return value
+    ///
+    /// Returns nothing on success, or an error string on failure.
+    pub fn write(&self, buffer: &mut Write) -> Result<(), String> {
+        let _ = match buffer.write_i32::<LittleEndian>(self.message_length) {
+            Ok(_) => (),
+            Err(_) => return Err("Unable to write message_length".to_owned())
+        };
```
kyeah: Since we're throwing out any result values, it'd be better to get rid of ```let _``` and ```match``` and do an if case on ```.is_err()```.

Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "mongodb"
 version = "0.1.0"
-authors = ["Valeri Karpov <valkar207@gmail.com>"]
+authors = ["Valeri Karpov <valkar207@gmail.com>",
+           "Saghm Rossi <saghmrossi@gmail.com",
```
kyeah: Missing closing braces on these emails.

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,309 @@
...
+            &Message::OpReply {..} => Err("OP_REPLY should not be sent by the client".to_owned()),
+            &Message::OpQuery {
+                header: ref h,
+                flags: f,
+                full_collection_name: ref fcn,
+                number_to_skip: ns,
+                number_to_return: nr,
+                query: ref q,
+                return_field_selector: ref rfs
+            } => Message::write_query(buffer, &h, f, &fcn, ns, nr, &q, &rfs)
```
kyeah: Can we pass the message into ```write_query()```? Looks funky to expand and send each component individually.

src/client/wire_protocol/header.rs
saghmrossi: I don't think the error is being thrown out; if you look at line 160 in operations.rs, the same error that's returned by Header#write is being returned from Message#write_query.

src/client/wire_protocol/operations.rs
jedestep: You're correct that if you want the Message to own its Query then you shouldn't pass as a reference. However this method is accessed as part of an API level operation and it generally is a bad experience to have an API call take ownership of your variables. Suppose I wrote the following program:

```
let foo = vec!["a", "b", "c"];
let mut query = Document::new();
query.insert("foo".to_string(), Bson::String("placeholder".to_string()));
for val in foo {
    query.set("foo", val);
    println!(mongo_client.find_one(query));
}
```
After the first `find_one` the query has been destroyed after the message object used to transmit it over the network fell out of scope (when the network I/O call returned most likely). This can be worked around by copying the query at each iteration, but personally I think that seems less desirable.

src/client/wire_protocol/operations.rs
saghmrossi: I definitely agree that it's not optimal, but since OpQuery is an enum variant, `write_query()` would have to repeat the pattern match on `self` to be able to access the fields, which I think is probably worse.

src/client/wire_protocol/operations.rs
saghmrossi: Kevin is in the process of writing an actual client implementation that will wrap the wire protocol interface, so ideally users of the driver wouldn't actually be directly calling any of the methods here. Would it make more sense to have the client taking the borrowed pointer and then calling `to_owned()` on it, or do you think it should be done here? I'm not really sure which way is more clean, so I'm open to suggestions on that.

vkarpov15: Either way, if the API takes ownership at some level, Jed's concern will be valid. I think @steveklabnik might be able to help resolve this particular thread.

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// `number_to_return - The total number of documents that should be returned by
+    ///                     the query.
+    /// `return_field_selector - An optional projection of which fields should be
+    ///                          present in the documents to be returned by the
+    ///                          query.
+    ///
+    /// # Return value
+    ///
+    /// Returns the newly-created Message.
+    pub fn new_query(header_request_id: i32, flags: i32,
```
steveklabnik: I would also comment with the same as I did in another pull request: this doesn't make a new query, it makes a new message, so I would call it `with_query`

src/client/wire_protocol/operations.rs
steveklabnik: So, my initial thought is basically the same as @saghmrossi 's: @jedestep 's concern is absolutely valid, you should only take ownership if you need it. And, depending, you may not _need_ it here, as @saghmrossi points out, you can call `.to_owned()` to get something owned, probably via a copy. I haven't looked at what `query` is yet, exactly...

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,312 @@
+use bson::Document as BSON;
```
steveklabnik: Types that are acronyms should be `Bson`, not `BSON`.

src/client/wire_protocol/operations.rs
steveklabnik: so, it appears to be coming from `brson-rs`, but the link in crates.io has _very_ old Rust code, I'm not sure how it even compiles...

src/client/wire_protocol/operations.rs
saghmrossi: Shoot, I must have misremembered the style guideline for that, for some reason I thought it was the other way around.

src/client/wire_protocol/operations.rs
saghmrossi: It looks like it's less than two weeks old to me: https://crates.io/crates/bson-rs

src/client/wire_protocol/header.rs
kyeah: Yeah, I'm talking about extraneous underscore matches. Everything except the error will be thrown out, so you can do ```if buffer.write_i32().is_err() { return Err("...") }```.

saghmrossi: Oh, I see, that makes sense.

src/client/wire_protocol/operations.rs
steveklabnik: oh ugh, i had the wrong tab open, you're correct

src/client/wire_protocol/operations.rs
```diff
@@ -0,0 +1,392 @@
...
+                number_to_return: nr,
+                query: ref q,
+                return_field_selector: ref rfs
+            } => Message::write_query(buffer, &h, &f, &fcn, ns, nr, &q, &rfs)
+        }
+    }
+
+    /// Reads a serialized reply message from a buffer
+    ///
+    /// FIXME: Right now, this returns only the first BSON document from the
```
vkarpov15: As a general rule of thumb, prefer opening up a github issue rather than 'TODO' or 'FIXME'. I have never seen somebody actually remove a commented out TODO from code.

saghmrossi: Should I take out the FIXME before the merge, or should I just fix it now? I only put that in because I had just gotten it working for that simple case and I wanted to make sure I got in the rest of the code before I left yesterday.

PR #43, opened 2015-06-10
Client API Base

kyeah:
Basic client with ```find``` and ```find_one``` on collections, and basic commands on databases. Skeleton included for other Query/Insert options and operations.

See tests for example client usage.

### DIFF
--- a/src/client/coll.rs
+++ b/src/client/coll.rs
@@ -1,0 +1,201 @@
+use bson;
+use client::db::Database;
+use client::common::{ReadPreference, WriteConcern};
+use client::wire_protocol::operations::{OpQueryFlags, Message};
+
+/// Interfaces with a MongoDB collection.
+pub struct Collection<'a> {
+    db: &'a Database<'a>,
+    pub name: String,
+    pub full_name: String,
+    read_preference: Option<ReadPreference>,
+    write_concern: Option<WriteConcern>,
+}
+
+#[derive(Clone, PartialEq)]
+pub enum CursorType {
+    NonTailable,
+    Tailable,
+    TailableAwait,
+}
+
+#[derive(Clone)]
+pub struct AggregateOptions {
+    pub allow_disk_use: bool,
+    pub use_cursor: bool,
+    pub batch_size: Option<i32>,
+    pub max_time_ms: Option<i64>,
+}
+
+#[derive(Clone)]
+pub struct CountOptions {
+    pub hint: Option<bson::Document>,
+    pub limit: Option<i64>,
+    pub max_time_ms: Option<i64>,
+    pub skip: Option<i64>,
+}
+
+#[derive(Clone)]
+pub struct DistinctOptions {
+    pub max_time_ms: Option<i64>,
+}
+
+#[derive(Clone)]
+pub struct FindOptions {
+    pub allow_partial_results: bool,
+    pub no_cursor_timeout: bool,
+    pub op_log_replay: bool,
+    pub skip: i32,
+    pub limit: i32,
+    pub cursor_type: CursorType,
+    pub batch_size: Option<i32>,
+    pub comment: Option<String>,
+    pub max_time_ms: Option<i64>,
+    pub modifiers: Option<bson::Document>,
+    pub projection: Option<bson::Document>,
+    pub sort: Option<bson::Document>,
+}
+
+impl FindOptions {
+    /// Creates a new FindOptions struct with default parameters.
+    pub fn new() -> FindOptions {
+        FindOptions {
+            allow_partial_results: false,
+            no_cursor_timeout: false,
+            op_log_replay: false,
+            skip: 0,
+            limit: 0,
+            cursor_type: CursorType::NonTailable,
+            batch_size: None,
+            comment: None,
+            max_time_ms: None,
+            modifiers: None,
+            projection: None,
+            sort: None,
+        }
+    }
+
+    pub fn with_skip(&self, skip: i32) -> FindOptions {
+        let mut new_opts = self.clone();
+        new_opts.skip = skip;
+        new_opts
+    }
+
+    pub fn with_limit(&self, limit: i32) -> FindOptions {
+        let mut new_opts = self.clone();
+        new_opts.limit = limit;
+        new_opts
+    }
+
+    pub fn with_projection(&self, proj: Option<bson::Document>) -> FindOptions {
+        let mut new_opts = self.clone();
+        new_opts.projection = proj;
+        new_opts
+    }
+}
+
+impl<'a> Collection<'a> {
+    /// Creates a collection representation with optional read and write controls.
+    ///
+    /// If `create` is specified, the collection will be explicitly created in the database.
+    pub fn new(db: &'a Database<'a>, name: &str, create: bool,
+               read_preference: Option<ReadPreference>, write_concern: Option<WriteConcern>) -> Collection<'a> {
+
+        let coll = Collection {
+            full_name: format!("{}.{}", db.name, name),
+            db: db,
+            name: name.to_owned(),
+            read_preference: read_preference,
+            write_concern: write_concern,
+        };
+
+        /*
+        // Since standard collections are implicitly created on insert,
+        // this should only be used to create capped collections.
+        if create {
+            coll.create();
+        };
+         */
+
+        coll
+    }
+
+    /// Returns a unique operational request id.
+    pub fn get_req_id(&self) -> i32 {
+        self.db.client.get_req_id()
+    }
+
+    // Read Spec
+    pub fn aggregate(pipeline: &[bson::Document], options: AggregateOptions) -> Result<Vec<bson::Document>, String> {
+        Err("IMPL".to_owned())
+    }
+
+    pub fn count(filter: bson::Document, options: CountOptions) -> Result<i64, String> {
+        Err("IMPL".to_owned())
+    }
+
+    pub fn distinct(field_name: &str, filter: bson::Document, options: DistinctOptions) -> Result<Vec<String>, String> {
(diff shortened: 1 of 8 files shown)


src/client/coll.rs
```diff
@@ -0,0 +1,201 @@
+use bson;
+use client::db::Database;
+use client::common::{ReadPreference, WriteConcern};
+use client::wire_protocol::operations::{OpQueryFlags, Message};
+
+/// Interfaces with a MongoDB collection.
+pub struct Collection<'a> {
+    db: &'a Database<'a>,
+    pub name: String,
+    pub full_name: String,
```
vkarpov15: This should be called 'namespace'. The 'dbname.collname' string is called a namespace in MongoDB parlance.

src/client/coll.rs
```diff
@@ -0,0 +1,201 @@
...
+            tailable_cursor: options.cursor_type != CursorType::NonTailable,
+            slave_ok: false,
+            oplog_relay: options.op_log_replay,
+            no_cursor_timeout: options.no_cursor_timeout,
+            await_data: options.cursor_type == CursorType::TailableAwait,
+            exhaust: false,
+            partial: options.allow_partial_results,
+        };
+
+        let req = try!(Message::with_query(self.get_req_id(), flags, self.full_name.to_owned(),
```
vkarpov15: In general, we should probably abstract out the wire protocol a little more here. Generally find() shouldn't interact with a socket directly.

src/client/mod.rs
```diff
@@ -1,2 +1,68 @@
...
+        Ok(MongoClient {
+            req_id: Cell::new(0),
+            socket: Arc::new(RefCell::new(socket)),
+            config: config,
+        })
+    }
+
+    /// Returns a unique operational request id.
+    pub fn get_req_id(&self) -> i32 {
+        self.req_id.set(self.req_id.get() + 1);
```
vkarpov15: Is this thread safe? I'm not terribly familiar with how cells work but this looks like a possible race condition.

tests/client/coll.rs
```diff
@@ -0,0 +1,50 @@
...
+use mongodb::client::MongoClient;
+use mongodb::client::db::Database;
+use mongodb::client::coll::{Collection, FindOptions};
+use std::io::Write;
+use std::net::TcpStream;
+
+//#[test]
+fn find() {
+    let client = MongoClient::with_uri("mongodb://localhost:27017").unwrap();
+    let db = Database::new(&client, "sample", None, None);
```
vkarpov15: In most drivers, db implements a `collection()` function that makes it so you don't have to do the `Collection::new` below. Makes for less boilerplate. Could we implement that?

src/client/coll.rs
kyeah: Sam has a working Cursor implementation, so the ```find``` methods should run through that once it's in the codebase.

src/client/coll.rs
```diff
@@ -0,0 +1,201 @@
...
+                                           options.skip, options.limit, doc, options.projection));
```
vkarpov15: In general, `number_to_return` is not the same as limit. In the cursor implementation, there's a separate concept of 'batch size' which is what you pass in instead, so you can set a large limit and the driver doesn't block until you get all the results back. This works for now but we'll have to re-address that when we implement cursors

src/client/mod.rs
kyeah: RefCell will panic! if two threads attempt to access the same socket at once using ```borrow_mut```. I'll add a mutex to prevent this from happening.

src/client/coll.rs
```diff
@@ -0,0 +1,201 @@
...
+            batch_size: None,
+            comment: None,
+            max_time_ms: None,
+            modifiers: None,
+            projection: None,
+            sort: None,
+        }
+    }
+
+    pub fn with_skip(&self, skip: i32) -> FindOptions {
```
saghmrossi: This method (as well as `with_projection`) could probably be changed to take `&mut self` as an argument instead of returning a cloned struct.

src/client/coll.rs
saghmrossi: I already have separate `batch_size` and `number_to_return` fields in my cursor implementation, so this won't be an issue once we combine our progress.

src/client/connstring.rs
```diff
@@ -86,7 +86,7 @@ impl ConnectionString {
     }
 }
 
-/// Parses a MongoDB connection string URI as defined by
+/// Parses a MongoDB connection string URI as defined by
```
saghmrossi: Trailing whitespace!

src/client/coll.rs
kyeah: I actually changed options to have public fields, so users can just set the option themselves if they wish. Removing ```with_skip``` and ```with_projection```, and keeping ```with_limit``` for ```find_one()``` until cursors are integrated.

src/client/mod.rs
kyeah: Oops, I was looking at the socket RefCell, which also needed fixing. This may be a race condition; looking into it.

src/client/coll.rs
```diff
@@ -0,0 +1,194 @@
+use bson;
+use client::db::Database;
+use client::common::{ReadPreference, WriteConcern};
+use client::wire_protocol::operations::{OpQueryFlags, Message};
+
+/// Interfaces with a MongoDB collection.
+pub struct Collection<'a> {
+    db: &'a Database<'a>,
+    pub name: String,
+    pub namespace: String,
```
alabid: It seems rather superfluous having "name" and "namespace" Strings represent a Collection. Within the server, we have a lightweight NamespaceString object (a full-qualified namespace for a collection like "test.first_coll" = <db>.<collection>). The name of a collection can easily be gotten from the NamespaceString.

src/client/coll.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct DistinctOptions {
+    pub max_time_ms: Option<i64>,
+}
+
+#[derive(Clone)]
+pub struct FindOptions {
+    pub allow_partial_results: bool,
+    pub no_cursor_timeout: bool,
+    pub op_log_replay: bool,
+    pub skip: i32,
```
alabid: `skip` can't be negative.

src/client/coll.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        let coll = Collection {
+            namespace: format!("{}.{}", db.name, name),
+            db: db,
+            name: name.to_owned(),
+            read_preference: read_preference,
+            write_concern: write_concern,
+        };
+
+        /*
```
alabid: remove this comment block

src/client/db.rs
```diff
@@ -0,0 +1,109 @@
...
+use bson::Bson;
+use client::MongoClient;
+use client::coll::Collection;
+use client::common::{ReadPreference, WriteConcern};
+
+/// Interfaces with a MongoDB database.
+pub struct Database<'a> {
+    pub name: String,
+    pub client: &'a MongoClient,
+    read_preference: Option<ReadPreference>,
```
alabid: Why are `read_preference` and `write_concern` members attached to the `Database`?  A write concern is enforced when doing writes (or when executing commands that could do writes such as write commands and findAndModify -- for 3.2 -- which I incidentally added to the server :) ). Think about what having a `Database` have a read preference and a write concern means. As an exercise, take a look at how other drivers use these fields and remove these fields for now. But I want you to start thinking about read preference and write concern.

tests/client/coll.rs
```diff
@@ -0,0 +1,50 @@
...
+//#[test]
+fn find() {
+    let client = MongoClient::with_uri("mongodb://localhost:27017").unwrap();
+    let db = client.db("sample");
+    let coll = db.collection("movies");
+    
+    let results = coll.find(None, None).unwrap();
+
+    assert!(results.len() > 0);
+    let expected_val = &Bson::String("Jaws".to_owned());
```
alabid: These tests are confusing. When is "Jaws" inserted into the "sample.movies" collection?

src/client/db.rs
kyeah: The [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#q-a) indicates that ```read_preference``` and ```write_concern``` should be specified on a Client/Database/Collection level, and both the python and c++ drivers follow this paradigm with the client also having these fields. The python driver also removed the ability to specify these at the operations level, while the c++ driver provides these as options for certain operations.

tests/client/coll.rs
kyeah: These tests (including the current ```wire_protocol``` tests) are all defunct since they rely on a custom local collection. Until inserts are supported, there isn't much use in having these tests in the codebase beyond being an example of library usage. Should I remove these for now?

src/client/db.rs
alabid: Alright. But we should allow also specifying the `write_concern` on a per-operation basis too. Remember to set `read_preference` and `write_concern` to their defaults. `read_preference`, pretty sure, defaults to `PRIMARY`.

src/client/coll.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct Collection<'a> {
+    db: &'a Database<'a>,
+    pub name: String,
+    pub namespace: String,
+    read_preference: Option<ReadPreference>,
+    write_concern: Option<WriteConcern>,
+}
+
+#[derive(Clone, PartialEq)]
+pub enum CursorType {
```
jedestep: Why is the equivalence only partial?

tests/client/coll.rs
vkarpov15: Nope keep them

PR #44, opened 2015-06-11
Insert functionality, refactored flags, and support for returning multiple documents from a query

saghmrossi:
Since I haven't commited all my refactoring for documentation/style, some of the documentation for these functions might be slightly out-of-date, but that should be with a new pull request relatively soon.


src/client/wire_protocol/flags.rs
```diff
@@ -0,0 +1,106 @@
...
+    cursor_not_found: bool, // Bit 0
+    query_failure: bool,    // Bit 1
+    await_capable: bool,    // Bit 3
+
+    // All bits remaining must be 0
+}
+
+impl OpReplyFlags {
+    pub fn from_i32(i: i32) -> OpReplyFlags {
+        let cnf = (i & 1) != 0;
```
vkarpov15: General naming convention: don't use acronyms unless they're well-known, like 'http' or 'bson'. Stick to 'cursor_not_found' as the var name

src/client/wire_protocol/operations.rs
```diff
@@ -306,6 +318,12 @@ impl Message {
         match self {
             /// Only the server should sent replies
             &Message::OpReply {..} => Err("OP_REPLY should not be sent by the client".to_owned()),
+            &Message::OpInsert {
+                header: ref h,
+                flags: ref f,
+                full_collection_name: ref fcn,
```
vkarpov15: ditto here, write out 'full_collection_name' instead of 'fcn', 'header' instead of 'h', etc.

vkarpov15:
Couple minor comments and looks like you need to merge upstream/1.0 back into your branch, there's a merge conflict

src/client/coll.rs
```diff
@@ -174,8 +175,14 @@ impl<'a> Collection<'a> {
         };
 
         try!(req.write(&mut *socket.borrow_mut()));
-        let bson = try!(Message::read(&mut *socket.borrow_mut()));
-        Ok(vec!(bson))
+        let message = try!(Message::read(&mut *socket.borrow_mut()));
+
+	match message {
```
alabid: fix indentation here

src/client/wire_protocol/flags.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl OpQueryFlags {
+    pub fn no_flags() -> OpQueryFlags {
+        OpQueryFlags { tailable_cursor: false, slave_ok: false,
+                       oplog_relay: false, no_cursor_timeout: false,
+                       await_data: false, exhaust: false, partial: false }
+    }
+
+    pub fn to_i32(&self) -> i32 {
```
alabid: I feel like this method is unnecessarily long. Why not just
```
if self.tailable_cursor {
    i |= (1 << 1)
}
...
```

src/client/coll.rs
saghmrossi: Oops, sorry about that

src/client/wire_protocol/flags.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut i = 0 as i32;
+
+        if self.tailable_cursor {
+            let bit = 1 << 1;
+
+            i |= bit;
+        }
+
+        if self.slave_ok {
+            let bit = 1 << 2;
```
alabid: And I didn't take a look at the original pull request to see that you were doing 
`i &= ...` instead

saghmrossi: Right, yeah, I noticed right after it got merged and asked Val if I should push again, but he said I could just group it with the next pull request.

src/client/wire_protocol/flags.rs
saghmrossi: Hmmm....I'm not actually sure how I ended up doing that in the way I did, but you're right, it's definitely unnecessary to have the temporary variable.

src/client/wire_protocol/operations.rs
```diff
@@ -380,10 +403,12 @@ impl Message {
...
 
         match header.op_code {
-            OpCode::OpReply => Message::read_reply(buffer),
+            OpCode::Reply => {
+                Message::read_reply(buffer, header)
+            },
             opcode => {
-                let s = format!("Expected to read response but instead found: {}", opcode.to_string());
-                Err(s)
+                let s = opcode.to_string();
```
alabid: you really don't need that variable

src/client/wire_protocol/flags.rs
```diff
@@ -0,0 +1,94 @@
+pub struct OpReplyFlags {
+    cursor_not_found: bool, // Bit 0
+    query_failure: bool,    // Bit 1
+    await_capable: bool,    // Bit 3
+
+    // All bits remaining must be 0
+}
+
```
alabid: put some documentation in thsi file

src/client/wire_protocol/operations.rs
saghmrossi: Without the temporary variable, the line is nearly 100 characters long, so I wanted to shorten it in some way to make it more readable. Would it be preferable to just use a backslash to split the string onto multiple lines?

alabid:
```
format!("Expected to read response but instead found: {}",
        opcode.to_string());
```

PR #45, opened 2015-06-12
run some real tests against mongodb 2.6 in travis

vkarpov15:
@kyeah @saghmrossi this will enable us to run tests in Travis until we get insert working.

The only test that's excluded for now is the `list_collections()` one, because list_collections got a big overhaul in 3.0 and our current implementation is incompatible with mongodb 2.6, which is fine. We'll figure out a way to use [mongodb-version-manager](https://www.npmjs.com/package/mongodb-version-manager) or something similar to make travis use mongodb 3.0 sometime next week.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
     - sudo make install
     - sudo ldconfig /usr/local/lib
 
-before_script: "echo \"db.test.insert({ foo: 42 }); db.getSiblingDB('sample').movies.insert({ title: 'Jaws' });\" | mongo"
+before_script: "db.getSiblingDB('sample').movies.insert({ title: 'Jaws' });\" | mongo"
 
 script:
     - cargo build --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: rust
 
+services:
+    - "mongodb"
+
 before_install:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,6 +12,8 @@ before_install:
     - sudo make install
     - sudo ldconfig /usr/local/lib
 
+before_script: "echo \"db.test.insert({ foo: 42 }); db.getSiblingDB('sample').movies.insert({ title: 'Jaws' });\" | mongo"
```
saghmrossi: What's the reasoning behind putting { foo: 42 } in the database before running the tests? Right now, the wire protocol tests assume that the database is empty initially.

vkarpov15: I haven't adjusted this for your outstanding PR yet. The code in 1.0 right now assumes that there's a `{ foo: 42 }` document in 'test.test'

saghmrossi: Ah, okay, makes sense

.travis.yml
```diff
@@ -9,6 +12,8 @@ before_install:
     - sudo make install
     - sudo ldconfig /usr/local/lib
 
+before_script: "db.getSiblingDB('sample').movies.insert({ title: 'Jaws' });\" | mongo"
```
kyeah: Missing `echo \"` before `db.getSiblingDB`, which is killing the travis build.

PR #46, opened 2015-06-15
Saghmrossi/command

saghmrossi:


saghmrossi:
FYI the travis build failed because there is no `mongod` server running; nothing actually got broken.

vkarpov15:
LGTM on my end. @alabid ?

saghmrossi:
This isn't actually supposed to be merged, since the command functionality is going to be directly placed in the client. Yesterday it was suggested that this be put into a pull request so Kevin could get an idea of how the code is structured, so that's why this is here.

PR #47, opened 2015-06-15
Saghmrossi/cursor

saghmrossi:


saghmrossi:
Cursor test failed because there isn't any `mongod` server running yet.

src/client/cursor.rs
```diff
@@ -0,0 +1,160 @@
...
+/// # Fields
+///
+/// `request_id` - Uniquely identifies the request being sent.
+/// `full_collection_name` - The full qualified name of the collection,
+///                          beginning with the database name and a period.
+/// `batch_size` - How many documents to fetch at a given time from the server.
+/// `cursor_id` - Uniquely identifies the cursor being returned by the reply.
+pub struct Cursor<'a, T> where T: Read + Write + 'a {
+    request_id: i32,
+    full_collection_name: String,
```
alabid: Please call this `namespace` to remain consistent. `Collection` has a `namespace` field.

src/client/cursor.rs
```diff
@@ -0,0 +1,160 @@
...
+
+                Some((v, cid))
+            },
+            _ => None
+        }
+    }
+
+    pub fn query_with_batch_size(stream: &'a mut T, batch_size: i32,
+                                 request_id: i32, flags: OpQueryFlags,
+                                 full_collection_name: &str,
```
alabid: `namespace` instead of `full_collection_name`

src/client/cursor.rs
```diff
@@ -0,0 +1,160 @@
...
+    ///
+    ///
+    fn get_bson_and_cid_from_message(message: Message) -> Option<(VecDeque<bson::Document>, i64)> {
+        match message {
+            Message::OpReply { header: _, flags: _, cursor_id: cid,
+                               starting_from: _, number_returned: _,
+                               documents: docs } => {
+                let mut v = VecDeque::new();
+
+                for bson in docs {
```
alabid: `bson`, in this case, represents a bson document. Change `bson` to something more appropriate.

src/client/cursor.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    pub fn next_batch(&mut self) -> Vec<bson::Document> {
+        let n = self.batch_size;
+
+        self.next_n(n)
+    }
+}
+
+impl <'a, T> Iterator for Cursor<'a, T> where T: Read + Write + 'a {
+    type Item = bson::Document;
```
alabid: unused alias.

src/client/wire_protocol/operations.rs
```diff
@@ -53,6 +53,13 @@ pub enum Message {
         query: BsonDocument,
         return_field_selector: Option<BsonDocument>,
     },
+    OpGetMore {
+        header: Header,
+        // ZERO goes here
```
alabid: I don't understand the `// ZERO goes here` comment.

src/client/cursor.rs
saghmrossi: Actually, this is needed to be able to implement the Iterator trait: http://rustbyexample.com/trait/iter.html

src/client/wire_protocol/operations.rs
saghmrossi: The wire protocol specifies that an empty int32 field goes there in the struct, so I thought it would be useful to use a comment as a placeholder to remind anyone reading the code that there needs to be an empty space there: http://docs.mongodb.org/meta-driver/latest/legacy/mongodb-wire-protocol/#wire-op-get-more

src/client/cursor.rs
alabid: Ok.

src/client/cursor.rs
saghmrossi: I'm not really sure I understand the issue with calling it this; I did the same thing in Message::with_insert.

src/client/wire_protocol/operations.rs
alabid: I'd prefer a comment that's more descriptive (like your reply that makes it clear that the ZERO field is for the wire protocol).

src/client/wire_protocol/operations.rs
```diff
@@ -289,6 +320,11 @@ impl Message {
         Ok(())
     }
 
+    pub fn write_get_more(buffer: &mut Write, header: &Header,
+                          full_collection_name: &str, number_to_return: i32,
+                          cursor_id: i64) -> Result<(), String> {
+        Err("".to_owned())
```
saghmrossi: Oops, definitely need to implement this before it gets merged.

tests/client/cursor.rs
```diff
@@ -0,0 +1,82 @@
...
+        Ok(mut stream) => {
+            let docs : Vec<_> = (0..10).map(|i| {
+                let mut doc = Document::new();
+                doc.insert("foo".to_owned(), I64(i));
+
+                doc
+            }).collect();
+
+            let flags = OpInsertFlags::no_flags();
+            let name = "test.test".to_owned();
```
vkarpov15: Would be better IMO for each test to use a separate collection name, just to minimize likelihood of stale state.

saghmrossi: Yeah, I guess that's probably a good idea until we have a way of dropping databases.

src/client/cursor.rs
```diff
@@ -0,0 +1,171 @@
...
+    ///
+    ///
+    fn get_bson_and_cid_from_message(message: Message) -> Option<(VecDeque<bson::Document>, i64)> {
+        match message {
+            Message::OpReply { header: _, flags: _, cursor_id: cid,
+                               starting_from: _, number_returned: _,
+                               documents: docs } => {
+                let mut v = VecDeque::new();
+
+                for bson_doc in docs {
```
vkarpov15: nitpick: `doc` is probably better than `bson_doc`, slightly redundant to name something `bson_doc` when that's the only type of document we're supporting for the foreseeable future.

src/client/cursor.rs
```diff
@@ -0,0 +1,171 @@
...
+            Some((buf, cursor_id)) => Ok(Cursor {
+                request_id: request_id,
+                namespace: namespace.to_owned(),
+                batch_size: batch_size,
+                cursor_id: cursor_id,
+                limit: number_to_return,
+                count: 0,
+                buffer: buf,
+                stream: stream }),
+            None => Err("Invalid resonse received".to_owned()),
```
vkarpov15: typo

src/client/cursor.rs
saghmrossi: Ugh, okay, I'll change it again

vkarpov15:
Couple minor comments.

src/client/cursor.rs
saghmrossi: Oops

PR #48, opened 2015-06-15
CRUD Spec and Basic Insert/Update/Delete write commands

kyeah:
This PR fills the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#) for collections, except for Errors, and implements the basic insert, update, and delete write commands without ObjectID generation. The coll module is now separated into `coll`, `coll::options`, and `coll::results`.

Other small updates (sorry):
- `db`: Added the `drop_collection()` function.
- `connstring`: Converted to using `BTreeMap` over `HashMap`. The standard `HashMap` implementation uses a secure crypto hash algorithm that makes it perform significantly slower than `BTreeMap`.
- `Cargo.toml`: Updated to use the insertion-order implementation of BSON documents, necessary for command ops.

The travis environment has been configured as per @saghmrossi's fix in #47 , but it will fail until we get 3.0 testing up and running.


src/client/coll/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            slave_ok: false,
+            oplog_relay: options.op_log_replay,
+            no_cursor_timeout: options.no_cursor_timeout,
+            await_data: options.cursor_type == CursorType::TailableAwait,
+            exhaust: false,
+            partial: options.allow_partial_results,
+        };
+
+        let req = try!(Message::with_query(self.get_req_id(), flags, self.namespace.to_owned(),
+                                           options.skip as i32, options.limit, doc, options.projection));
```
saghmrossi: Is there any reason why `skip` isn't being stored as an int32 in all of the options structs?

kyeah: `skip` can't be negative (as per @alabid), but the server still takes it as i32. `u32` prevents it on the client-side. However, perhaps it should be stored as `i32` to follow spec?

saghmrossi: What about in CountOptions, where it's 64-bits?

kyeah: Same question for unsigned; 64-bit count options are specified in the spec.

src/client/coll/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let options = match options {
+            Some(opts) => opts,
+            None => FindOptions::new(),
+        };
+
+        let res = try!(self.find(filter, Some(options.with_limit(1))));
+        match res.len() {
+            0 => Ok(None),
+            _ => Ok(Some(res[0].to_owned())),
```
alabid: We should error if we get more than one document when we specify `limit: 1`

src/client/coll/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            None => Err("delete_many reply not received from server.".to_owned()),
+        }
+    }
+
+    /// Replaces a single document.
+    pub fn replace_one(&self, filter: bson::Document, replacement: bson::Document, upsert: bool,
+                       write_concern: Option<WriteConcern>) -> Result<UpdateResult, String> {
+
+        for key in replacement.keys() {
+            if key.starts_with("$") {
```
alabid: Not sure how necessary front-end checks when the server will always error out when `replacement.keys()` contains disallowed keys.

src/client/coll/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        unimplemented!()
+    }
+
+    /// Finds the distinct values for a specified field across a single collection.
+    pub fn distinct(field_name: &str, filter: bson::Document, options: DistinctOptions) -> Result<Vec<Bson>, String> {
+        unimplemented!()
+    }
+
+    /// Returns a list of documents within the collection that match the filter.
+    pub fn find(&self, filter: Option<bson::Document>, options: Option<FindOptions>)
```
alabid: `find` should return a cursor. For now, we can leave this as is so we keep moving forward.
But put a
`TODO: <information about returning a cursor instead of a vector of documents`
and try to resolve the TODO soon.

kyeah: Yep, this was next on my to-do list. I'll open an issue since Val doesn't like TODO comments :)

alabid: sounds good.

src/client/coll/mod.rs
kyeah: Seems like a fairly harmless check. This is also the only thing that differentiates replace and update, so it may prevent mix-ups if someone uses `update` and accidentally provides a replace document.

src/client/coll/mod.rs
vkarpov15: Thanks @kyeah :)

src/client/coll/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            None => bson::Document::new(),
+        };
+
+        let options = match options {
+            Some(opts) => opts,
+            None => FindOptions::new(),
+        };
+
+        let flags = OpQueryFlags {
+            tailable_cursor: options.cursor_type != CursorType::NonTailable,
```
vkarpov15: We should really have a function to abstract away this logic and the logic on L105. Would be handy to have a constructor that abstracts away this != and == stuff

vkarpov15:
Couple small comments. Excellent work overall!

@steveklabnik, comments on the general rustyness of this code would be much appreciated. This implements a lot of the [basic functionality we want out of a mongodb driver](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#) - inserts, updates, etc.

src/client/coll/mod.rs
alabid: ++

vkarpov15:
Sweet. Can you fix the merge conflict please @kyeah ?

steveklabnik:
I will check out the 'general rustiness' soon :)

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    /// Returns a unique operational request id.
+    pub fn get_req_id(&self) -> i32 {
+        self.db.client.get_req_id()
+    }
+
+    /// Extracts the collection name from the namespace.
+    pub fn name(&self) -> String {
+        match self.namespace.find(".") {
+            Some(idx) => self.namespace[idx+1..].to_owned(),
```
steveklabnik: careful with this. This is a byte index of these strings, are you guaranteed that this is actually correct? Can namespaces have non-ascii characters?

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /// Returns a list of documents within the collection that match the filter.
+    pub fn find(&self, filter: Option<bson::Document>, options: Option<FindOptions>)
+                -> Result<Vec<bson::Document>, String> {
+
+        let doc = match filter {
+            Some(bson) => bson,
+            None => bson::Document::new(),
+        };
```
steveklabnik: This code pattern happens a lot here. You might be interested in http://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unwrap_or , which would change this line to

    let doc = filter.unwrap_or(bson::Document::new())

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        let flags = OpQueryFlags::with_find_options(&options);
+        let req = try!(Message::with_query(self.get_req_id(), flags, self.namespace.to_owned(),
+                                           options.skip as i32, options.limit, doc, options.projection));
+
+        let socket = match self.db.client.socket.lock() {
+            Ok(val) => val,
+            _ => return Err("Client socket lock poisoned.".to_owned()),
+        };
```
steveklabnik: Bummer we can't make `try` variadric over `Option` too :(

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        let res = try!(self.find(filter, Some(options.with_limit(1))));
+        match res.len() {
+            0 => Ok(None),
+            1 => Ok(Some(res[0].to_owned())),
+            n => Err(format!("Expected a single document, found {}.", n)),
+        }
+    }
+
+    // Find and Modify Spec
```
steveklabnik: given that the next line is a doc comment, this should probably be removed

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        let res = try!(self.db.command(cmd));
+        match res {
+            Some(doc) => Ok(doc),
+            None => Err("Insertion reply not received from server.".to_owned()),
+        }
+    }
+
+    /// Inserts the provided document. If the document is missing an identifier,
+    ///  the driver should generate one.
```
steveklabnik: leading space here

src/client/coll/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /// Finds a single document and updates it, returning either the original
+    ///  or updated document.
+    pub fn find_one_and_update(&self, filter: bson::Document, update: bson::Document,
+                               options: Option<FindOneAndUpdateOptions>)  -> Option<bson::Document> {
+        unimplemented!()
+    }
+
+    // Write Spec
```
steveklabnik: same here, this should probably go

src/client/coll/results.rs
```diff
@@ -0,0 +1,104 @@
+use bson;
+use bson::Bson;
+
+use std::collections::BTreeMap;
+
+/// Results for a bulk write operation.
+#[derive(Clone)]
+pub struct BulkWriteResult {
```
steveklabnik: Since `Result` means something in Rust, this is a bit odd. I'm not sure if that's enough to figure out a better name, and I'm not sure what name to suggest...

src/client/coll/mod.rs
kyeah: What would be the best way to grab a non-ascii substring? http://is.gd/PIX31l19 (via Rust forum)?

src/client/coll/results.rs
kyeah: Possibly `Outcome`, although I'm wary of moving away from the naming conventions in the mongo driver spec.

vkarpov15: I think we'll keep 'Result', that's something that the CRUD spec isn't very flexible on unfortunately

tests/client/coll.rs
```diff
@@ -1,50 +1,303 @@
...
+
+    set.insert("director".to_owned(), Bson::String("Robert Zemeckis".to_owned()));
+    update.insert("$set".to_owned(), Bson::Document(set));
+    coll.update_many(doc2.clone(), update, false, None).ok().expect("Failed to update documents.");
+
+    let results = coll.find(None, None).ok().expect("Failed to execute find command.");
+    assert_eq!(4, results.len());
+
+    // Assert director attributes
+    for i in 0..3 {
```
vkarpov15: Test assertions in for loops are very difficult to debug because the line number doesn't exactly correspond to the assertion error (doesn't tell you which value of loop variable the tests failed). Can you flatten this out? As a general rule of thumb, `for` and `if` should not be in test code.

vkarpov15: @kyeah can you address this in a separate PR? Don't want to block y'all anymore on this.

kyeah: Yep, working on it now and there's a bug in the `update_many` test, will open a new PR.

PR #52, opened 2015-06-16
Travis test

saghmrossi:
Don't merge this, I'm just testing travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: rust
 
+services:
+    - "mongodb"
+
 script:
     - cargo build --verbose
     - export RUST_TEST_THREADS=1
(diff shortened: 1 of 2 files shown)


saghmrossi:
@vkarpov15  Looks like travis is passing all the tests in 1.0 with the mongo 2.6 service. What are my next steps?

vkarpov15:
I'll merge this. I think the best way to go forward would be to pull down https://www.npmjs.com/package/mongodb-version-manager and use it to start mongodb 3.0. I would guess you should be able to install nodejs in travis using something like [this](https://github.com/mongodb-js/connect-mongodb-session/blob/master/.travis.yml#L2-3)

PR #54, opened 2015-06-16
changed travis to use direct download of mongodb 3.0

saghmrossi:
Testing travis with mongodb 3.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ before_script:
     - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.0.3.tgz
     - tar xvf mongodb-linux-x86_64-3.0.3.tgz
     - mv mongodb-linux-x86_64-3.0.3 3.0.3
-    - mkdir -p /data/db
 
 env:
     - RUST_TEST_THREADS=1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,15 @@
...
-services:
-    - "mongodb"
+before_script:
+    - wget https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.0.3.tgz
+    - tar xvf mongodb-linux-x86_64-3.0.3.tgz
+    - mv mongodb-linux-x86_64-3.0.3 3.0.3
+    - mkdir -p /data/db
+
+env:
+  - RUST_TEST_THREADS=1
```
vkarpov15: Nitpick: consistent spacing

saghmrossi: Oops, sorry, I'm used to two-spaces for YAML, but the first travis commit used four, so I keep forgetting to change mine.

vkarpov15:
Agreed re: 4 spaces vs 2 for yaml. We may want to tweak that at some point, but since 4 spaces seems to be the rust standard and yaml doesn't care it's not that big of a deal.

PR #56, opened 2015-06-16
Integrate cursors into find queries

### LINKED ISSUE #50: collection.find() should return Cursors
As of #48, collection.find() returns a list of bson documents. The expected behavior should be to return a lazy-eval cursor to iterate over batches.

kyeah:
Modifies cursors to contain Collection refs. `find` and `find_one` now return cursors.

Fixes #50.

### DIFF
--- a/src/client/coll/mod.rs
+++ b/src/client/coll/mod.rs
@@ -9,12 +9,14 @@ use client::common::{ReadPreference, WriteConcern};
 use client::coll::options::*;
 use client::coll::results::*;
 
+use client::cursor::Cursor;
+
 use client::wire_protocol::flags::OpQueryFlags;
 use client::wire_protocol::operations::Message;
 
 /// Interfaces with a MongoDB collection.
 pub struct Collection<'a> {
-    db: &'a Database<'a>,
+    pub db: &'a Database<'a>,
     pub namespace: String,
     read_preference: ReadPreference,
     write_concern: WriteConcern,
@@ -62,7 +64,7 @@ impl<'a> Collection<'a> {
     }
 
     /// Runs an aggregation framework pipeline.
-    pub fn aggregate(pipeline: &[bson::Document], options: AggregateOptions) -> Result<Vec<bson::Document>, String> {
+    pub fn aggregate(pipeline: &[bson::Document], options: AggregateOptions) -> Result<Cursor<'a>, String> {
         unimplemented!()
     }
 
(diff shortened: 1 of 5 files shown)


src/client/cursor.rs
```diff
@@ -136,13 +126,12 @@ impl <'a, T> Cursor<'a, T> where T: Read + Write + 'a {
...
-                 number_to_return: i32, query: bson::Document,
-                 return_field_selector: Option<bson::Document>) -> Result<Cursor<'a, T>, String> {
-
-        Cursor::query_with_batch_size(stream, 20, request_id, flags,
-                                      namespace, number_to_skip,
+    pub fn query(collection: &'a Collection<'a>, flags: OpQueryFlags,
+                 number_to_skip: i32, number_to_return: i32, query: bson::Document,
+                 return_field_selector: Option<bson::Document>) -> Result<Cursor<'a>, String> {
+        
+        Cursor::query_with_batch_size(collection, 20, flags,
```
vkarpov15: magic number

kyeah: Will replace with a DEFAULT_BATCH_SIZE const.

src/client/coll/mod.rs
```diff
@@ -77,30 +79,15 @@ impl<'a> Collection<'a> {
...
-        try!(req.write(&mut *socket.borrow_mut()));
-        let message = try!(Message::read(&mut *socket.borrow_mut()));
-
-        match message {
-            Message::OpReply { header: _, flags: _, cursor_id: _,
-                               starting_from: _, number_returned: _,
-                               documents, } => Ok(documents),
-            _ => Err("Invalid response received from server".to_owned())
-        }
+        Cursor::query_with_batch_size(self, options.batch_size, flags, options.skip as i32,
```
vkarpov15: I think we need to special case `options.limit = 1` somewhere, because with [OpQuery, when `number_to_return` is 1, the server assumes its a find_one and kills the cursor](http://docs.mongodb.org/meta-driver/latest/legacy/mongodb-wire-protocol/#op-query)

src/client/coll/options.rs
```diff
@@ -128,7 +128,7 @@ impl FindOptions {
             skip: 0,
             limit: 0,
             cursor_type: CursorType::NonTailable,
-            batch_size: None,
+            batch_size: 20,
```
vkarpov15: let's hide 20 behind a constant somewhere re: previous comment

alabid: ++. Define the default batch size in `client/cursor.rs`

src/client/cursor.rs
```diff
@@ -14,18 +18,16 @@ use std::io::{Read, Write};
...
     batch_size: i32,
     cursor_id: i64,
     limit: i32,
     count: i32,
     buffer: VecDeque<bson::Document>,
-    stream: &'a mut T,
 }
 
-impl <'a, T> Cursor<'a, T> where T: Read + Write + 'a {
+impl <'a> Cursor<'a> {
```
alabid: I think the cursor should have `has_next` and `next` methods. The shell has `next()` and `hasNext()` methods on the cursor.

kyeah: There is a `next()` method (implemented by @saghmrossi for Iterator). I'll add `has_next()`.

src/client/coll/mod.rs
kyeah: Added a check in `has_next()`.

PR #57, opened 2015-06-17
Macros

saghmrossi:


tests/client/coll.rs
```diff
@@ -85,17 +94,25 @@ fn list_collections() {
...
+    };
 
-    doc.insert("title".to_owned(), Bson::String("Jaws".to_owned()));
-    doc2.insert("title".to_owned(), Bson::String("Back to the Future".to_owned()));
-    coll.insert_many(vec!(doc, doc2), false, None).ok().expect("Failed to insert documents.");
+    let doc2 = doc! {
+        "title" => Bson::String("Back to the Future".to_owned())
+    };
+
+    // let mut doc = bson::Document::new();
```
alabid: why is this commented out and not removed?

tests/client/coll.rs
```diff
@@ -7,13 +8,17 @@ use mongodb::client::MongoClient;
...
     db.drop_database().ok().expect("Failed to drop database");
 
     // Insert document
-    let mut doc = bson::Document::new();
-    doc.insert("title".to_owned(), Bson::String("Jaws".to_owned()));
+    let doc = doc! {
+        "title" => Bson::String("Jaws".to_owned())
+    };
+
+    // let mut doc = bson::Document::new();
```
vkarpov15: Another good rule to follow: don't leave dead code hanging around in comments. Once code is commented out, it's difficult for people later on to figure out whether its safe to remove, so it often just lives on in the code base forever. If you want to find what the state of the code was previously, that's what the git history is for.

vkarpov15:
lgtm modulo details with commented out code

vkarpov15:
Also, @steveklabnik, this is our first attempt at using macros for building up bson docs. It's admittedly naive but works well thus far. Do you have any suggestions for improvements?

steveklabnik:
I am really bad at macros but seems fine. You might want to look at what other people have done for hashmaps... for example, https://github.com/rust-lang/rust/issues/14726

PR #58, opened 2015-06-17
Basic RunCommands

kyeah:
Implements command cursors and modifies the base Cursor implementation to borrow the MongoClient instead of the Collection. Also implements `is_master()` and `database_names()` for clients.

The travis config has been modified to run a clean server on port 27018 under `data/test` for testing `database_names()`.

Successor to #46.

### DIFF
--- a/tests/client/client.rs
+++ b/tests/client/client.rs
@@ -16,7 +16,7 @@ fn list_databases() {
 #[test]
 fn database_names() {
     let client = MongoClient::with_uri("mongodb://localhost:27018").unwrap();
-    let mut results = client.database_names().ok().expect("Failed to execute database_names.");
+    let results = client.database_names().ok().expect("Failed to execute database_names.");
     assert_eq!(1, results.len());
     assert_eq!("local", results[0]);
 }
--- a/tests/client/client.rs
+++ b/tests/client/client.rs
@@ -1,0 +1,21 @@
+use bson::Bson;
+use mongodb::client::MongoClient;
+
+#[test]
+fn list_databases() {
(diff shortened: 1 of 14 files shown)


tests/client/client.rs
```diff
@@ -0,0 +1,21 @@
+use bson::Bson;
+use mongodb::client::MongoClient;
+
+#[test]
+fn list_databases() {
```
alabid: Can you add more tests for getting a cursor for `list_databases`?
Test that you can get a cursor with `results < batch size, results > batchSize, and results == batchSize`

kyeah: On a second look at the python driver and the cmd reply, it looks like `list_databases` returns all databases in a single document without a cursor_id, so it should be run through `command` and return a `Vec<String>`. It may be more API consistent to return a cursor and provide a separate `database_names()` function, though, since `list_collections()` definitely provides some form of batching using a cursor.

vkarpov15: That behavior is very much version dependent, you may be looking at an older version of the python driver or using mongodb <= 2.6. For instance, [the node driver explicitly checks if the server has a list collections command, and returns a cursor if it does](https://github.com/mongodb/node-mongodb-native/blob/47dbb499283ba9887bc7602a743d4719f6332b46/lib/db.js#L477-488).

kyeah: `list_databases()` removed in favor of `database_names()`, which returns the array of db names. `list_collections()` still uses a cursor (assuming 3.0+ for now).

kyeah:
Note that the `database_names()` test is destructive in that it clears the server of all databases except for `local`.

PR #59, opened 2015-06-18
Query and Write Op Commands

kyeah:
Implements `aggregate()`, `count()`, `distinct()`, and all of the FindAndModify commands on collections with preliminary testing (extensive spec test suite being built by @saghmrossi).

### DIFF
--- a/src/client/coll/mod.rs
+++ b/src/client/coll/mod.rs
@@ -96,6 +96,8 @@ impl<'a> Collection<'a> {
         if filter.is_some() {
             spec.insert("query".to_owned(), Bson::Document(filter.unwrap()));
         }
+
+        // Favor specified hint document over string
         if opts.hint_doc.is_some() {
             spec.insert("hint".to_owned(), Bson::Document(opts.hint_doc.unwrap()));
         } else if opts.hint.is_some() {
--- a/src/client/coll/mod.rs
+++ b/src/client/coll/mod.rs
@@ -101,24 +101,76 @@ impl<'a> Collection<'a> {
         Ok(cursor.next())
     }
 
+    fn find_one_and_modify(&self, cmd: &mut bson::Document,
+                           filter: bson::Document, max_time_ms: Option<i64>,
+                           projection: Option<bson::Document>, sort: Option<bson::Document>)
+                           -> Result<Option<bson::Document>, String> {
+
+        cmd.insert("findAndModify".to_owned(), Bson::String(self.name()));
+        cmd.insert("query".to_owned(), Bson::Document(filter));
+        if sort.is_some() {
+            cmd.insert("sort".to_owned(), Bson::Document(sort.unwrap()));
+        }
+        if projection.is_some() {
+            cmd.insert("fields".to_owned(), Bson::Document(projection.unwrap()));
+        }
+
+        let res = try!(self.db.command(cmd.to_owned()));
+        match res {
+            Some(doc) => match doc.get("value") {
+                Some(&Bson::Document(ref nested_doc)) => Ok(Some(nested_doc.to_owned())),
+                _ => Ok(None),
+            },
+            None => Ok(None),
+        }
+    }
+
+    fn find_and_update(&self, filter: bson::Document, update: bson::Document,
+                       after: bool, max_time_ms: Option<i64>, projection: Option<bson::Document>,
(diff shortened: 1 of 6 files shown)


src/client/coll/mod.rs
```diff
@@ -101,24 +160,81 @@ impl<'a> Collection<'a> {
         Ok(cursor.next())
     }
 
+    fn find_one_and_modify(&self, cmd: &mut bson::Document,
```
alabid: Just call this method `findAndModify`. You can't find many and modify anyways.

src/client/coll/mod.rs
```diff
@@ -101,24 +160,81 @@ impl<'a> Collection<'a> {
...
+        match res {
+            Some(doc) => match doc.get("value") {
+                Some(&Bson::Document(ref nested_doc)) => Ok(Some(nested_doc.to_owned())),
+                _ => Ok(None),
+            },
+            None => Ok(None),
+        }
+    }
+
+    fn find_and_update(&self, filter: bson::Document, update: bson::Document,
```
vkarpov15: I'd recommend renaming this as `find_and_modify` - `find_and_update` is not very mongodb-ish nomenclature. Even if it's private, we should stick to more consistent naming conventions

src/client/coll/mod.rs
vkarpov15: +1

tests/client/coll.rs
```diff
@@ -59,22 +53,308 @@ fn find_and_insert_one() {
...
+    } };
+
+    // Aggregate
+    let mut cursor = coll.aggregate(vec![project, unwind, group], None)
+        .ok().expect("Failed to execute aggregate command.");
+
+    let results = cursor.next_n(10);
+    assert_eq!(6, results.len());
+
+    for i in 0..6 {
```
vkarpov15: Let's not do complex for loops in tests, that just makes test failures very difficult to debug

src/client/coll/mod.rs
```diff
@@ -64,18 +64,77 @@ impl<'a> Collection<'a> {
...
+        if opts.hint_doc.is_some() {
+            spec.insert("hint".to_owned(), Bson::Document(opts.hint_doc.unwrap()));
+        } else if opts.hint.is_some() {
+            spec.insert("hint".to_owned(), Bson::String(opts.hint.unwrap()));
+        }
+
+        let result = try!(self.db.command(spec));
+        if result.is_some() {
+            match result.unwrap().get("n") {
+                Some(&Bson::I32(ref n)) => return Ok(*n as i64),
```
alabid: In what cases would we get `I32` vs `I64`?

kyeah: The specs indicate that `count()` returns i64's, but the command provides i32's. I matched the i64 case as a precaution, as it's unclear to me why the discrepancy exists and whether there were planned changes to the server to start sending i64, whether it sends i32 for small document counts, etc.

src/client/coll/mod.rs
```diff
@@ -101,24 +160,83 @@ impl<'a> Collection<'a> {
         Ok(cursor.next())
     }
 
+    // Helper method for all findAndModify commands.
+    fn find_and_modify(&self, cmd: &mut bson::Document,
```
alabid: `findAndModify` now takes a write concern.

kyeah: From [findAndModify docs](http://docs.mongodb.org/manual/reference/command/findAndModify/):
> You cannot specify a write concern to findAndModify to override the default write concern whereas, starting in MongoDB 2.6, you can specify a write concern to the update() method.

Not sure why the command doesn't allow a specified WriteConcern.

src/client/coll/options.rs
```diff
@@ -119,6 +120,40 @@ pub struct FindOneAndUpdateOptions {
     pub upsert: bool,
 }
 
+impl AggregateOptions {
+    pub fn new() -> AggregateOptions {
+        AggregateOptions {
+            allow_disk_use: false,
+            use_cursor: true,
```
alabid: Is the default to use a cursor?

kyeah: yep. CRUD spec:

> Indicates whether the command will request that the server provide results using a cursor.  The default is true.

PR #60, opened 2015-06-22
Change macros to be recursive

saghmrossi:
I figured out a way to make the `doc!` macro recursive (so we won't have to use `nested_doc!` for nested documents I also added in some syntactic sugar making BSON array literals as values in documents. The main trick to making this possible was having the macro parse the values as token trees rather than expressions, which has the slight downside of requiring values that aren't arrays or documents to be wrapped in parentheses, but I think this is definitely worth it to get rid of `nested_doc! { ... } ` and `Bson::Array(vec![...]))` in favor of `{ ... }` and `[ ... ]`.

I put a pretty comprehensive example in `tests/macros/mod.rs`, along with some helper functions (and a macro) to print out BSON documents in a more human-readable way. If you want to run the test in Cargo to see what the output looks like, you'll need to run it with `cargo test recursive_macro -- --nocapture` (the `-- --nocapture` allows the tests to print to standard output, which otherwise won't occur).

I'm submitting this as a pull request mostly just as a convenient way to show the code, but I'm wondering if this functionality should be put in a pull request to the BSON library that we're using. The helper functions used to print out the BSON documents could pretty easily be refactored into a ToString implementation, and it also makes some sense to bundle the macros with the actual definition of the BSON data structure rather than in this libary (although we'd want to make a point in our documentation that these exist to make sure that users of the driver are aware of them).

### DIFF
--- a/tests/client/coll.rs
+++ b/tests/client/coll.rs
@@ -1,6 +1,4 @@
-use bson;
-use bson::Bson;
-use bson::Document;
+use bson::{Bson, Document};
 
 use mongodb::client::MongoClient;
 
@@ -14,7 +12,7 @@ fn find_and_insert() {
 
     // Insert document
     let doc = doc! {
-        "title" => Bson::String("Jaws".to_owned())
+        "title" => (Bson::String("Jaws".to_owned()))
     };
 
     coll.insert_one(doc, None).ok().expect("Failed to insert document");
@@ -42,7 +40,7 @@ fn find_and_insert_one() {
 
     // Insert document
     let doc = doc! {
-        "title" => Bson::String("Jaws".to_owned())
+        "title" => (Bson::String("Jaws".to_owned()))
     };
 
     coll.insert_one(doc, None).ok().expect("Failed to insert document");
@@ -68,11 +66,11 @@ fn insert_many() {
 
     // Insert documents
     let doc1 = doc! {
-        "title" => Bson::String("Jaws".to_owned())
+        "title" => (Bson::String("Jaws".to_owned()))
     };
 
     let doc2 = doc! {
-        "title" => Bson::String("Back to the Future".to_owned())
+        "title" => (Bson::String("Back to the Future".to_owned()))
     };
 
     coll.insert_many(vec![doc1, doc2], false, None).ok().expect("Failed to insert documents.");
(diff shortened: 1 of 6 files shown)


tests/macros/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+
+#[test]
+fn recursive_macro() {
+    let doc = doc! {
+        "a" => (Bson::String("foo".to_owned())),
+        "b" => {
+            "bar" => {
+                "harbor" => [Bson::String("seal".to_owned()), Bson::Boolean(false)],
+                "jelly" => (Bson::FloatingPoint(42.0))
```
vkarpov15: image

tests/client/coll.rs
```diff
@@ -14,7 +12,7 @@ fn find_and_insert() {
 
     // Insert document
     let doc = doc! {
-        "title" => Bson::String("Jaws".to_owned())
+        "title" => (Bson::String("Jaws".to_owned()))
```
vkarpov15: Any way to work around having to put the parens here or can we use a different delimiter? Parenthesis are cool but the fact that parenthesis matter here makes the code a little more difficult to read. Might even be better to have an explicit `v!()` macro.

saghmrossi: I can change it to work with any delimeter, but it just has to be different than the braces and brackets used for the arrays and nested documents.

Alternately, I could change it to use a different symbol after (or instead of) "=>", but it would have to be before the actual value to get rid of the parsing ambiguity. (i.e. something like `"key" => $value` could work, with the dollar sign or some other token resolving the ambiguity).

saghmrossi: Or how about adding an extra `>` after the arrow (for single values only, not arrays or docs), so it would look like `"key" =>> value`?

saghmrossi: Actually, from re-examining it, this might be trickier to change than I thought. I'll have to spend some time to see if getting rid of the parentheses is possible.

vkarpov15: `$` would probably be sufficient actually. I'm just concerned that parenthesis would add a lot of confusion because that overrides how parenthesis work in Rust. If we could get something like `"title" => $Bson::String("Jaws".to_owned())` that would actually be a good alternative too.

saghmrossi: I'm not sure that could work in the current way the macro is implemented, but I could try to rewrite it again in a way that can do that.

vkarpov15: So let's merge it this way for now, and see if there's a way we can improve it later. This way is miles ahead of our current mechanisms for building bson docs :)

PR #61, opened 2015-06-22
Automated tests for the CRUD spec ("find" and "insert_one")

saghmrossi:


saghmrossi:
@vkarpov15 Looks like the cloning of the "specifications" repo is working.

tests/json/eq.rs
```diff
@@ -0,0 +1,91 @@
...
+trait NumEq {
+    fn float_eq(&self, f: f64) -> bool;
+    fn int_eq(&self, i: i64) -> bool;
+}
+
+impl NumEq for Bson {
+    fn float_eq(&self, f: f64) -> bool {
+        match self {
+            &Bson::FloatingPoint(ref ff) => f.approx_eq(ff),
+            &Bson::I32(i) => f == (i as f64),
```
alabid: should we use `approx_eq` here too?

saghmrossi: Oops, yeah

tests/client/crud_spec/framework.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        assert!(eq::bson_eq(&id, &inserted));
+        check_coll!($db, $c, $o.collection);
+    }};
+}
+
+
+#[macro_export]
+macro_rules! run_suite {
+    ( $f:expr, $c:expr ) => {{
```
vkarpov15: Let's avoid 1-char variable names, even for macros. Could we write `$file` instead?

tests/lib.rs
```diff
@@ -2,6 +2,9 @@ extern crate bson;
 
 #[macro_use(add_to_doc, doc)]
 extern crate mongodb;
+extern crate rustc_serialize;
+extern crate nalgebra as na;
```
vkarpov15: Don't rename packages unless necessary - makes it trickier to dig out what 'na' means.

saghmrossi: Okay, I was just following the example in the README:

---
The preferred way to use nalgebra is to import types and traits explicitly, and call free-functions using the na:: prefix:

```
extern crate nalgebra as na;
use na::{Vec3, Rot3, Rotation};

fn main() {
    let     a = Vec3::new(1.0f64, 1.0, 1.0);
    let mut b = Rot3::new(na::zero());

    b.append_rotation_mut(&a);

    assert!(na::approx_eq(&na::rotation(&b), &a));
}
```
---
https://github.com/sebcrozet/nalgebra

tests/client/crud_spec/read.rs
```diff
@@ -0,0 +1,12 @@
+use bson::Bson;
+use json::arguments::Arguments;
+use json::reader::SuiteContainer;
+use json::eq;
+use mongodb::client:: MongoClient;
+use rustc_serialize::json::Json;
+
+#[test]
+fn find() {
+    run_suite!("tests/json/data/specs/source/crud/tests/read/find.json",
```
alabid: nice.

PR #62, opened 2015-06-22
Removing some of the warnings

saghmrossi:
Most of the warnings are caused by struct fields  and the parameters to the unimplemented methods in the API not being  used, but a few legitimate ones had fallen through the cracks. This pull request removes the ones that can be safely removed.


vkarpov15:
lgtm

PR #64, opened 2015-06-23
Mongo Errors

kyeah:
This commit introduces a library-wide `mongodb::Error` enum and associated `mongodb::Result` to provide more versatile error logging and handling. In addition, Cursor network errors are now accounted for. `WriteError` and `BulkWriteError` are provided as defined in the CRUD specifications, but are not yet handled accordingly in write operations.

Most of the vital changes/additions are in `client/error.rs` and `client/cursor.rs`.


src/client/cursor.rs
```diff
@@ -58,46 +59,42 @@ impl <'a> Cursor<'a> {
...
 
-    fn get_bson_and_cid_from_command_message(message: Message) -> Option<(VecDeque<bson::Document>, i64, String)> {
-        match Cursor::get_bson_and_cid_from_message(message) {
-            Some((v, _)) => {
-                if v.len() != 1 {
-                    return None;
-                }
-
-                let ref doc = v[0];
+    fn get_bson_and_cid_from_command_message(message: Message) -> Result<(VecDeque<bson::Document>, i64, String)> {
```
alabid: should the return value be `Result<(VecDeque<bson::Document>, i64)>`?

src/client/error.rs
```diff
@@ -0,0 +1,264 @@
...
+
+use client::coll::options::WriteModel;
+use std::{error, fmt, io, result, sync};
+
+/// A type for results generated by MongoDB related functions, where the Err type is mongodb::Error.
+pub type Result<T> = result::Result<T, Error>;
+
+/// The error type for MongoDB operations.
+#[derive(Debug)]
+pub enum Error {
```
alabid: Put comments about what each error type means. It's not clear what `CursorMissingError` should represent, for example.

src/client/error.rs
```diff
@@ -0,0 +1,264 @@
...
+pub struct WriteException {
+    kind: WriteExceptionKind,
+    code: i32,
+    message: String,
+}
+
+/// The two types of single-write related MongoDB operations.
+#[derive(Debug)]
+pub enum WriteExceptionKind {
+    WriteConcernError(bson::Document),
```
alabid: A `writeConcernError` can also happen in multiple-write scenarios, for example, when executing write commands.

alabid: Also, a `WriteConcernError` is pretty different from a `WriteError`. For example, `findAndModify` could succeed (on the primary) but the writes could fail to be replicated resulting in a `WriteError`.

src/client/error.rs
```diff
@@ -0,0 +1,264 @@
...
+pub enum Error {
+    IoError(io::Error),
+    EncoderError(bson::EncoderError),
+    DecoderError(bson::DecoderError),
+    WriteError(WriteException),
+    BulkWriteError(BulkWriteException),
+    ArgumentError(String),
+    OperationError(String),
+    ResponseError(String),
+    CursorMissingError,
```
alabid: I don't really like the name of this error. `CursorNotFoundError` might be more appropriate.

tests/client/coll.rs
```diff
@@ -410,7 +447,11 @@ fn delete_many() {
     // Delete document
     coll.delete_many(doc2.clone(), None).ok().expect("Failed to delete documents.");
     let mut cursor = coll.find(None, None).ok().expect("Failed to execute find command.");
-    let result = cursor.next().unwrap();
+    let result = match cursor.next() {
+        Some(Ok(res)) => res,
+        Some(Err(_)) => panic!("Received error from 'cursor.next()'."),
+        None => panic!("Expected bson."),
```
alabid: Are we going to reach this pattern? And in what cases?

src/client/wire_protocol/operations.rs
```diff
@@ -662,23 +528,15 @@ impl Message {
...
+    pub fn read<T>(buffer: &mut T) -> Result<Message> where T: Read + Write {
+        let header = try!(Header::read(buffer));
         match header.op_code {
             OpCode::Reply => {
                 Message::read_reply(buffer, header)
             },
             opcode => {
-                Err(format!("Expected to read response but instead found: {}",
-                            opcode.to_string()))
+                Err(ResponseError(format!("Expected to read response but instead found opcode {}",
```
alabid: If you're going to include information about the specific opcode, you might as well also include `OpCode::Reply` in the error message.

src/client/error.rs
```diff
@@ -0,0 +1,264 @@
...
+        match self {
+            &Error::WriteError(ref inner) => inner.fmt(fmt),
+            &Error::BulkWriteError(ref inner) => inner.fmt(fmt),
+            &Error::EncoderError(ref inner) => inner.fmt(fmt),
+            &Error::DecoderError(ref inner) => inner.fmt(fmt),
+            &Error::IoError(ref inner) => inner.fmt(fmt),
+            &Error::ArgumentError(ref inner) => inner.fmt(fmt),
+            &Error::OperationError(ref inner) => inner.fmt(fmt),
+            &Error::ResponseError(ref inner) => inner.fmt(fmt),
+            &Error::CursorMissingError => write!(fmt, "No cursor found"),
```
alabid: Let's standardize our error message reporting. Sometimes we end with period and sometimes don't.

src/client/cursor.rs
kyeah: The function needs to also return the command namespace, so this is correct. I'll rename this function to make it more accurate.

tests/client/coll.rs
kyeah: Err occurs when the cursor fails to retrieve more results from the server (via network and encoding/decoding error). Some(Ok()) and None() are the standard iterator pattern matches for more docs or none left. 

The iterator trait has an `Option<T>` return for `next()`, which is why the pattern is a little funky, as the Result has to be nested in.

src/client/error.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// An invalid function or operational argument was provided.
+    ArgumentError(String),
+    /// A database operation failed to send or receive a reply.
+    OperationError(String),
+    /// A database operation returned an invalid reply.
+    ResponseError(String),
+    /// A cursor operation failed to return a cursor.
+    CursorNotFoundError,
+    /// The application failed to secure the client connection socket due to a poisoned lock.
+    LockError,
```
alabid: Why not `PoisonLockError`? LockError looks too generic.

PR #68, opened 2015-06-24
Use BSON Conversion Traits

kyeah:
I've updated the BSON library to provide `From<T>` conversion traits to Bson, allowing the `doc!` macro to automagically convert raw formats into most types of Bson (except for `JavaScriptCode(String)` and `Timestamp(i64)`, whose inner objects overlap with the common String and I64 types). This reduces boilerplate code considerably, allowing us to convert code like this:

```rust
let doc1 = doc! { "tags" => [Bson::String("a".to_owned()),
                             Bson::String("b".to_owned()),
                             Bson::String("c".to_owned())] };

let doc2 = doc! { "tags" => [Bson::String("a".to_owned()),
                             Bson::String("b".to_owned()),
                             Bson::String("d".to_owned())] };

let doc3 = doc! { "tags" => [Bson::String("d".to_owned()),
                             Bson::String("e".to_owned()),
                             Bson::String("f".to_owned())] };
```

into this:

```rust
let doc1 = doc! { "tags" => ["a"', "b", "c"] };
let doc2 = doc! { "tags" => ["a", "b", "d"] };
let doc3 = doc! { "tags" => ["d", "e", "f"] };
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,16 +2,16 @@
 name = "mongodb"
 version = "0.1.0"
 dependencies = [
- "bson 0.1.0 (git+https://github.com/kyeah/bson-rs)",
+ "bson 0.1.2 (git+https://github.com/kyeah/bson-rs)",
  "byteorder 0.3.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "nalgebra 0.2.17 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


vkarpov15:
Travis build failed for some reason, are you sure the gitspec is right? Restarting it.

kyeah:
Looks like I updated to the wrong bson-rs commit. Should work now.

PR #70, opened 2015-06-24
Automated tests for the CRUD spec ("replace_one")

saghmrossi:
This should only be merged **after** the pull request for the `update` commands since this is branched off from that.


saghmrossi:
Assuming that there aren't any issues with the code, this can be merged now since the `update` pull request is merged.

PR #75, opened 2015-06-26
OID Generation

kyeah:
Provides an implementation for client-side generation of ObjectIds. Interfaces with `libc` are necessary for retrieving the machine and process ids, introducing a couple of `unsafe` blocks.


tests/client/oid.rs
```diff
@@ -0,0 +1,57 @@
...
+    let s = "541b1a00e8a23afa832b218e";
+    let oid_res = oid::with_string(s);
+    assert!(oid_res.is_ok());
+    let oid = oid_res.unwrap();
+    let bytes: [u8; 12] = [0x54u8, 0x1Bu8, 0x1Au8, 0x00u8,
+                           0xE8u8, 0xA2u8, 0x3Au8, 0xFAu8,
+                           0x83u8, 0x2Bu8, 0x21u8, 0x8Eu8];
+
+    for i in 0..12 {
+        assert_eq!(bytes[i], oid[i]);
```
alabid: `oid` should implement the `core::cmp::Eq` trait. There are many situations when a user would want to compare object ids.

tests/client/oid.rs
```diff
@@ -0,0 +1,57 @@
+use mongodb::client::oid;
+use rustc_serialize::hex::ToHex;
+
```
alabid: I want to see some test cases testing endianness. For example in the server, we have: https://github.com/mongodb/mongo/blob/master/src/mongo/bson/oid_test.cpp#L77

src/client/oid.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    let mut buf: [u8; 8] = [0; 8];
+    BigEndian::write_u64(&mut buf, u_int);
+
+    let buf_u24: [u8; 3] = [buf[0], buf[1], buf[2]];
+    Ok(buf_u24)
+}
+
+#[test]
+fn pid_generation() {
```
alabid: why is this test here and not in `tests/client/oid.rs`?

tests/client/oid.rs
kyeah: oops, this can actually be compared directly (including inequalities).

src/client/oid.rs
```diff
@@ -0,0 +1,216 @@
...
+    // Init oid counter
+    if OID_COUNTER.load(Ordering::SeqCst) == 0 {
+        let mut rng = try!(OsRng::new());
+        let start = rng.gen_range(0, MAX_U24 + 1);
+        OID_COUNTER.store(start, Ordering::SeqCst);
+    }
+
+    let u_counter = OID_COUNTER.fetch_add(1, Ordering::SeqCst);
+
+    // Mod result instead of OID_COUNTER to prevent threading issues.
```
alabid: cool

src/client/oid.rs
kyeah: These are unit tests that rely on private functions, so they can't be isolated into a separate test module.

src/client/oid.rs
```diff
@@ -0,0 +1,216 @@
...
+    let mut buf: [u8; 4] = [0; 4];
+    for i in 0..3 {
+        buf[i] = oid[4+i];
+    }
+    LittleEndian::read_u32(&buf)
+}
+
+/// Retrieves the process id associated with an ObjectId.
+pub fn get_pid(oid: [u8; 12]) -> u16 {
+    LittleEndian::read_u16(&oid[7..])
```
alabid: I don't want to see "magic" number around. Define special numbers as constant. For example: `kTimestampOffset`, `kInstanceUniqueOffset`, and `kIncrementOffset` instantiated here: thttps://github.com/mongodb/mongo/blob/master/src/mongo/bson/oid.cpp#L47

src/client/oid.rs
```diff
@@ -0,0 +1,216 @@
...
+/// for more information.
+pub fn generate() -> Result<[u8; 12]> {
+    let timestamp = gen_timestamp();
+    let machine_id = try!(gen_machine_id());
+    let process_id = gen_pid();
+    let counter = try!(gen_count());
+
+    let mut buf: [u8; 12] = [0; 12];
+    for i in 0..4 { buf[i] = timestamp[i]; }
+    for i in 0..3 { buf[4+i] = machine_id[i]; }
```
alabid: Don't use magic numbers.

Cargo.lock
```diff
@@ -4,8 +4,12 @@ version = "0.1.0"
 dependencies = [
  "bson 0.1.2 (git+https://github.com/kyeah/bson-rs)",
  "byteorder 0.3.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
```
alabid: I like the way the dependencies here are listed lexicographically. Maybe we should do the same when using crates? What does the rust style guide say about that?

src/client/oid.rs
alabid: Ok.

Cargo.lock
kyeah: The guide doesn't seem to say anything explicitly, but they should be listed lexicographically; looks like that's the case in Cargo.toml and lib.rs. Are you talking about module-level imports (`use ...`)?

alabid: yep.

saghmrossi: This is a generally good idea, but in a couple edges cases we'd have to not to that in order for things to work. The main one I'm thinking of is when a module with macros is imported in lib.rs/mod.rs; the macros will only be accessible for the modules declared *after* the import/declaration of the module with macros, so those ones would have to appear first.

PR #79, opened 2015-06-30
Initial work to support bulk writes

saghm:
The initial implementations of the data structures pertaining to bulk writing were written well before the implementation of the actual functionality, so some minor changes and additions had to be made to make things easier. I also added the functionality for InsertOne models and tests for it.

### DIFF
--- a/tests/client/bulk.rs
+++ b/tests/client/bulk.rs
@@ -23,15 +23,15 @@ fn bulk_ordered_insert_only() {
 
     for (i, result) in cursor.into_iter().enumerate() {
         let doc = result.unwrap();
-        let i = i + 1;
+        let expected_id = i + 1;
 
         match doc.get("_id") {
-            Some(&Bson::I32(j)) => assert_eq!(i as i32, j),
+            Some(&Bson::I32(j)) => assert_eq!(expected_id as i32, j),
             _ => panic!("Invalid id: {:?}", doc)
         }
 
         match doc.get("x") {
-            Some(&Bson::I32(j)) => assert_eq!(11 * i as i32, j),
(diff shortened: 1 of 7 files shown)


tests/client/bulk.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    coll.bulk_write(models, true);
+
+    let cursor : Vec<_> = coll.find(None, None).unwrap().collect();
+
+    assert_eq!(cursor.len(), 4);
+
+    for (i, result) in cursor.into_iter().enumerate() {
+        let doc = result.unwrap();
+        let i = i + 1;
```
alabid: shouldn't `i` correspond to the `_id` field? Why are we incrementing it?

saghm: Since `i` is the index, it starts at 0, but the `_id` fields start at one. I just figured the easiest way to deal with this was to add one to it beforehand rather than using `i + 1` twice, since the need to case in both cases and the need to multiply in the second would make the additional operation a bit messy.

alabid: ah. I see. Let's have another variable called `expected_id` or something you deem more appropriate and clearer.

tests/client/bulk.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        match doc.get("_id") {
+            Some(&Bson::I32(j)) => assert_eq!(i as i32, j),
```
alabid: I think you forgot to update this particular test

saghm: Oops

saghm:
I had to add the structs in for Kevin to be able to write the functions I need for the rest of the bulk operations.

PR #81, opened 2015-07-01
Bulk delete and update ops

kyeah:
Implements private bulk update and delete operations on collections.


saghm:
FYI I'm currently blocked on this @alabid

src/client/coll/mod.rs
```diff
@@ -323,7 +324,7 @@ impl<'a> Collection<'a> {
                                                   exception)
                 } else {
                     self.execute_insert_many_batch(documents, ordered,
-                                                       result, exception)
```
alabid: good catch.

PR #83, opened 2015-07-02
Connection pooling using Condvar

kyeah:
Connections are now explicitly taken and returned to the pool, similar to how pools are implemented in hyper. To allow pools to be capped on the number of open connections, a condition variable is introduced for locking the connection pool; if all available connections are in use, the thread will release the pool lock until it is notified that a connection has been returned.


alabid:
We talked offline about a couple of small issues on Friday and you've resolved those issues.

PR #84, opened 2015-07-02
Finished bulk write functionality

saghm:
The test for ordered bulk write is passing. I didn't have a chance to write one for unordered bulk writes, but I still thought I should get the pull request in before the long weekend.

### DIFF
--- a/src/client/coll/error.rs
+++ b/src/client/coll/error.rs
@@ -274,7 +274,16 @@ impl BulkWriteException {
         }
     }
 
-    pub fn add_bulk_write_exception(&mut self, exception: BulkWriteException) {
+    pub fn add_bulk_write_exception(&mut self, exception_opt: Option<BulkWriteException>,
+                                    models: Vec<WriteModel>) -> bool {
+        let exception = match exception_opt {
+            Some(exception) => exception,
+            None => {
+                self.processed_requests.extend(models.into_iter());
+                return true
+            }
+        };
+
         for req in exception.processed_requests.iter() {
             self.processed_requests.push(req.clone());
         }
@@ -292,25 +301,27 @@ impl BulkWriteException {
         };
 
         self.message.push_str(&exception.message);
+
+        false
     }
 
-    pub fn add_write_exception(&mut self, exception: WriteException, i: i32,
-                               req: WriteModel) {
-        self.add_unproccessed_model(req.clone());
-
-        if exception.write_concern_error.is_some() {
-            self.write_concern_error = exception.write_concern_error;
-        };
-
-        if let Some(write_err) = exception.write_error {
-            let bulk_write_error = BulkWriteError::new(i, write_err.code,
-                                                       write_err.message,
-                                                       Some(req));
-            self.write_errors.push(bulk_write_error);
-        }
-
-        self.message.push_str(&exception.message);
-    }
+    // pub fn add_write_exception(&mut self, exception: WriteException, i: i32,
+    //                            req: WriteModel) {
+    //     self.add_unproccessed_model(req.clone());
+    //
+    //     if exception.write_concern_error.is_some() {
+    //         self.write_concern_error = exception.write_concern_error;
+    //     };
+    //
+    //     if let Some(write_err) = exception.write_error {
+    //         let bulk_write_error = BulkWriteError::new(i, write_err.code,
+    //                                                    write_err.message,
+    //                                                    Some(req));
+    //         self.write_errors.push(bulk_write_error);
+    //     }
+    //
+    //     self.message.push_str(&exception.message);
+    // }
 
     /// Validates a bulk write result.
     pub fn validate_bulk_write_result(result: bson::Document, write_concern: WriteConcern) -> Result<()> {
--- a/src/client/coll/error.rs
+++ b/src/client/coll/error.rs
@@ -305,24 +305,6 @@ impl BulkWriteException {
         false
     }
 
-    // pub fn add_write_exception(&mut self, exception: WriteException, i: i32,
-    //                            req: WriteModel) {
-    //     self.add_unproccessed_model(req.clone());
-    //
-    //     if exception.write_concern_error.is_some() {
-    //         self.write_concern_error = exception.write_concern_error;
-    //     };
-    //
-    //     if let Some(write_err) = exception.write_error {
-    //         let bulk_write_error = BulkWriteError::new(i, write_err.code,
-    //                                                    write_err.message,
-    //                                                    Some(req));
-    //         self.write_errors.push(bulk_write_error);
-    //     }
-    //
-    //     self.message.push_str(&exception.message);
-    // }
-
     /// Validates a bulk write result.
     pub fn validate_bulk_write_result(result: bson::Document, write_concern: WriteConcern) -> Result<()> {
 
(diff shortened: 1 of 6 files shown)


src/client/coll/error.rs
```diff
@@ -292,26 +301,28 @@ impl BulkWriteException {
...
-                                                       write_err.message,
-                                                       Some(req));
-            self.write_errors.push(bulk_write_error);
-        }
-
-        self.message.push_str(&exception.message);
+        false
     }
 
+    // pub fn add_write_exception(&mut self, exception: WriteException, i: i32,
```
alabid: why is this commented out and not removed?

src/client/coll/mod.rs
```diff
@@ -171,7 +171,7 @@ impl<'a> Collection<'a> {
 
     // Helper method for all findAndModify commands.
     fn find_and_modify(&self, cmd: &mut bson::Document,
-                       filter: bson::Document, max_time_ms: Option<i64>,
+                       filter: bson::Document, _max_time_ms: Option<i64>,
```
alabid: why are you renaming to `_max_time_ms`?

src/client/coll/error.rs
saghm: Oops, forgot to push the change where I deleted this.

src/client/coll/mod.rs
```diff
@@ -254,95 +254,165 @@ impl<'a> Collection<'a> {
...
-        for req in requests {
-            match req {
-                WriteModel::InsertOne { document }  => {
-                    inserts.push(document)
-                }
-                _ => ()
+        let mut batches = vec![Batch::from(first_model)];
+
+        for model in requests {
+            let i = batches.len() - 1;
```
alabid: this variable `i` is not just used for iteration. Name appropriately.

src/client/coll/mod.rs
saghm: Didn't meant to leave this in, I just did this to demonstrate to Kevin how to get rid of the `unused variable` warning. I'll put this in a separate pull request with other refactors.

src/client/coll/mod.rs
```diff
@@ -254,95 +254,165 @@ impl<'a> Collection<'a> {
...
     }
 
     /// Sends a batch of writes to the server at the same time.
     pub fn bulk_write(&self, requests: Vec<WriteModel>, ordered: bool) -> BulkWriteResult {
         let batches = if ordered {
-            Collection::get_ordered_batches(requests)
-        } else {
-            Collection::get_unordered_batches(requests)
-        };
+                          Collection::get_ordered_batches(VecDeque::from_iter(requests.into_iter()))
```
alabid: I feel like we're going back-and-forth on this style. Kevin's last merged PR aligned the if/else block with the let. Talk to Kevin.

saghm: I checked the official Rust repo, and it seems that the unindented style is what they use, so I'll stick with that (see line 530 of https://github.com/rust-lang/rust/blob/f76d9bcfc2c269452522fbbe19f66fe653325646/src/libsyntax/ext/build.rs)

tests/client/bulk.rs
```diff
@@ -71,3 +71,142 @@ fn bulk_unordered_insert_only() {
...
+
+    assert_eq!(result.inserted_count, 9);
+    assert_eq!(result.inserted_ids.len() as i32, result.inserted_count);
+    assert_eq!(result.matched_count, 3);
+    assert_eq!(result.modified_count, 3);
+    assert_eq!(result.deleted_count, 3);
+    assert_eq!(result.upserted_count, 1);
+    assert_eq!(result.upserted_ids.len() as i32, result.upserted_count);
+
+    match result.inserted_ids.get(&0).unwrap() {
```
alabid: Lines 153-211 can be condensed to <20 lines.

tests/client/bulk.rs
```diff
@@ -71,3 +71,103 @@ fn bulk_unordered_insert_only() {
...
+
+    assert_eq!(result.inserted_count, 9);
+    assert_eq!(result.inserted_ids.len() as i32, result.inserted_count);
+    assert_eq!(result.matched_count, 3);
+    assert_eq!(result.modified_count, 3);
+    assert_eq!(result.deleted_count, 3);
+    assert_eq!(result.upserted_count, 1);
+    assert_eq!(result.upserted_ids.len() as i32, result.upserted_count);
+
+    macro_rules! check_value_in_tree {
```
alabid: We don't need to know that `inserted_ids` is a BTree map.

src/client/coll/results.rs
```diff
@@ -87,39 +87,81 @@ impl BulkWriteResult {
...
+    }
+
+    fn parse_upserted_ids(bson: Bson, start_index: i64,
+                          upserted_ids: &mut BTreeMap<i64, Bson>) -> i32 {
+        match bson {
+            Bson::Document(doc) => BulkWriteResult::parse_upserted_id(doc, start_index, upserted_ids),
+            Bson::Array(vec) => {
+                let mut count = 0;
+
+                for bson in vec {
```
alabid: rename `bson` variable here.

src/client/coll/mod.rs
```diff
@@ -38,7 +38,7 @@ impl<'a> Collection<'a> {
     /// Creates a collection representation with optional read and write controls.
     ///
     /// If `create` is specified, the collection will be explicitly created in the database.
-    pub fn new(db: &'a Database<'a>, name: &str, create: bool,
+    pub fn new(db: &'a Database<'a>, name: &str, _create: bool,
```
alabid: Why is `_create` still here?

src/client/coll/mod.rs
```diff
@@ -127,7 +127,7 @@ impl<'a> Collection<'a> {
     /// Finds the distinct values for a specified field across a single collection.
     pub fn distinct(&self, field_name: &str, filter: Option<bson::Document>, options: Option<DistinctOptions>) -> Result<Vec<Bson>> {
 
-        let opts = options.unwrap_or(DistinctOptions::new());
+        let _opts = options.unwrap_or(DistinctOptions::new());
```
alabid: should you remove `_opts`?

kyeah: The only option for `distinct` is maxTimeMs, which doesn't seem to be supported in the current op spec. Looking at the new `find` spec drafts, I assume this is for future 3.2+ support; the underscore hides the warning and allows us to easily add support later (standard rust style).

src/client/coll/mod.rs
kyeah: Pymongo provides the same option; we just haven't implemented `create_collection` yet in db.

alabid: ok

src/client/coll/mod.rs
alabid: ok

tests/client/bulk.rs
saghm: What exactly do you want changed here? Just the macro name? It seems rather inconsequential, given that it's local to the function and therefore won't be used elsewhere.

src/client/coll/results.rs
saghm: Not really sure what else to call it, given that it's some arbitrary bson value; there's no way to know if it's a document, array, number, or anything else.

src/client/coll/mod.rs
saghm: Sorry, meant for this to be in a separate pull request also.

src/client/coll/results.rs
alabid: ok

PR #85, opened 2015-07-06
Library Refactoring and Organization

kyeah:
Refactoring and organization that should be done before digging into our dogfooding projects (web app + gridfs).

* Made `client` module the root module
* Renamed MongoClient to `mongodb::Client`
* Moved ObjectId and macros to bson-rs
* Updated tests to reflect updated macro
* Updated README


saghm:
@alabid I'm blocked on this since  Cargo seems to pull in an incompatible `bson` version when specifying this as a dependency right now

PR #87, opened 2015-07-07
Fixed the "borrowing issue"

saghm:
Instead of `Database` having a reference to `Client` and `Collection` having a reference to `Database`, they now both use `Arc`'s, which allows them to be used in multiple threads.

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -34,7 +34,8 @@ impl Collection {
     /// Creates a collection representation with optional read and write controls.
     ///
     /// If `create` is specified, the collection will be explicitly created in the database.
-    pub fn new(db: Database, name: &str, _create: bool, read_preference: Option<ReadPreference>,
+    pub fn new(db: Database, name: &str, _create: bool,
+               read_preference: Option<ReadPreference>,
                write_concern: Option<WriteConcern>) -> Collection {
 
         let rp = read_preference.unwrap_or(db.read_preference.to_owned());
@@ -95,7 +96,8 @@ impl Collection {
     }
 
     /// Gets the number of documents matching the filter.
-    pub fn count(&self, filter: Option<bson::Document>, options: Option<CountOptions>) -> Result<i64> {
+    pub fn count(&self, filter: Option<bson::Document>,
+                 options: Option<CountOptions>) -> Result<i64> {
         let opts = options.unwrap_or(CountOptions::new());
 
         let mut spec = bson::Document::new();
@@ -122,8 +124,8 @@ impl Collection {
     }
 
     /// Finds the distinct values for a specified field across a single collection.
-    pub fn distinct(&self, field_name: &str, filter: Option<bson::Document>, options: Option<DistinctOptions>) -> Result<Vec<Bson>> {
-
+    pub fn distinct(&self, field_name: &str, filter: Option<bson::Document>,
+                    options: Option<DistinctOptions>) -> Result<Vec<Bson>> {
         let _opts = options.unwrap_or(DistinctOptions::new());
 
         let mut spec = bson::Document::new();
@@ -141,8 +143,8 @@ impl Collection {
     }
 
     /// Returns a list of documents within the collection that match the filter.
-    pub fn find(&self, filter: Option<bson::Document>, options: Option<FindOptions>)
-                -> Result<Cursor> {
+    pub fn find(&self, filter: Option<bson::Document>,
+                options: Option<FindOptions>) -> Result<Cursor> {
 
         let doc = filter.unwrap_or(bson::Document::new());
         let options = options.unwrap_or(FindOptions::new());
@@ -155,8 +157,8 @@ impl Collection {
     }
 
     /// Returns the first document within the collection that matches the filter, or None.
-    pub fn find_one(&self, filter: Option<bson::Document>, options: Option<FindOptions>)
-                    -> Result<Option<bson::Document>> {
+    pub fn find_one(&self, filter: Option<bson::Document>,
+                    options: Option<FindOptions>) -> Result<Option<bson::Document>> {
         let options = options.unwrap_or(FindOptions::new());
         let mut cursor = try!(self.find(filter, Some(options.with_limit(1))));
         match cursor.next() {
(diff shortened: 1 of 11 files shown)


alabid:
Great idea!

alabid:
As a follow-up to this perhaps, I'd love to see some tests particularly for the multi-threading functionality. @saghm thoughts?

saghm:
That probably could be done. The main reason that I made these changes was that they're necessary to be able to reuse the client over multiple requests in the webapp.

alabid:
Ok

src/coll/mod.rs
```diff
@@ -10,9 +10,10 @@ use self::error::{BulkWriteException, WriteException};
 use self::options::*;
 use self::results::*;
 
+use {ThreadedClient};
```
alabid: do we need the braces here?

src/cursor.rs
```diff
@@ -121,12 +121,12 @@ impl <'a> Cursor<'a> {
     ///
     /// Returns the cursor for the query results on success, or an Error on
     /// failure.
-    pub fn query_with_batch_size<'b>(client: &'a Client,
+    pub fn query_with_batch_size(client: Client,
                                      namespace: String, batch_size: i32,
```
alabid: fix indentation here.

kyeah:
Just thought of this: Would it be better to get rid of the Inner classes and have the structs be self-referential? i.e. have Client with `arc: Option<Arc<Client>>` and Database with `arc: Option<Arc<Database>>`, and instantiate them as so:

```rust
pub fn new(...) -> Arc<Client> {
    let mut client = Client { 
        arc: None,
        ...
    };

    let arc = Arc::new(client);
    client.arc = Some(arc.clone());
    arc
}
```

This would remove the need for auxiliary structs and the use of large additional traits that need to be defined and then imported by the user.

saghm:
I'm not sure that would be a cleaner solution; it seems kind of messy to just have an infinitely recursive reference in a struct, even if it does get rid of a couple of imports.

src/coll/mod.rs
saghm: Oops

kyeah:
Also seems messy to implement an entire struct/class within a public trait. ¯\_(ツ)_/¯

saghm:
There's no other way to add methods to an object type (i.e. `Arc`) that already exists. That's a limitation of Rust rather than the code itself.

kyeah:
Yup, just mentioning that both have their downsides.

saghm:
I definitely think that this way of doing things shouldn't be a permanent solution, but we definitely need some sort of multithreading support for the driver to be worth using. Maybe we could take another look at this again after the demo? We're already a bit tight for time (or at least, I am with the webapp, and I need some sort of multithreading support to be able to use the driver at all).

kyeah:
For sure! Not saying this shouldn't be merged. Just thought this would be a good place to bring up another possible option down the line that wouldn't be too much work to implement. :+1:

saghm:
Let's open an issue for this. It's definitely worth keeping the discussion open beyond the pull request.

src/db.rs
```diff
@@ -1,58 +1,82 @@
...
+    fn collection_names(&self, filter: Option<bson::Document>) -> Result<Vec<String>>;
+    fn create_collection(&self, name: &str) -> Result<()>;
+    fn drop_database(&self) -> Result<()>;
+    fn drop_collection(&self, name: &str) -> Result<()>;
+}
+
+impl ThreadedDatabase for Database {
+    /// Creates a database representation with optional read and write controls.
+    fn open(client: Client, name: &str, read_preference: Option<ReadPreference>,
+               write_concern: Option<WriteConcern>) -> Database {
```
alabid: pls. fix indentation here

src/db.rs
```diff
@@ -1,58 +1,82 @@
...
+    fn collection(&self, coll_name: &str) -> Collection {
+        Collection::new(self.clone(), coll_name, false, Some(self.read_preference.to_owned()), Some(self.write_concern.to_owned()))
     }
 
     /// Creates a collection representation with custom read and write controls.
-    pub fn collection_with_prefs(&'a self, coll_name: &str, create: bool,
-                                 read_preference: Option<ReadPreference>, write_concern: Option<WriteConcern>) -> Collection<'a> {
-        Collection::new(self, coll_name, create, read_preference, write_concern)
+    fn collection_with_prefs(&self, coll_name: &str, create: bool,
+                                 read_preference: Option<ReadPreference>,
```
alabid: also fix indentation here.

src/lib.rs
```diff
@@ -55,19 +74,19 @@ impl Client {
...
-    pub fn with_uri_and_prefs(uri: &str, read_pref: Option<ReadPreference>,
+    fn with_uri_and_prefs(uri: &str, read_pref: Option<ReadPreference>,
                               write_concern: Option<WriteConcern>) -> Result<Client> {
         let config = try!(connstring::parse(uri));
         Client::with_config(config, read_pref, write_concern)
     }
 
     fn with_config(config: ConnectionString, read_pref: Option<ReadPreference>,
-                   write_concern: Option<WriteConcern>) -> Result<Client> {
+        write_concern: Option<WriteConcern>) -> Result<Client> {
```
alabid: Are you using an editor different from the one you use normally?

src/lib.rs
```diff
@@ -55,19 +74,19 @@ impl Client {
...
     /// [the manual](http://docs.mongodb.org/manual/reference/connection-string/).
-    pub fn with_uri(uri: &str) -> Result<Client> {
+    fn with_uri(uri: &str) -> Result<Client> {
         Client::with_uri_and_prefs(uri, None, None)
     }
 
     /// `with_uri` with custom read and write controls.
-    pub fn with_uri_and_prefs(uri: &str, read_pref: Option<ReadPreference>,
+    fn with_uri_and_prefs(uri: &str, read_pref: Option<ReadPreference>,
                               write_concern: Option<WriteConcern>) -> Result<Client> {
```
alabid: ++

alabid:
Good idea opening the github issue. We need better multi-threading support. But I'll merge this PR once the indentation issues are fixed (and no other problems are introduced).

src/lib.rs
saghm: Sorry, no, it's just that I when I went through and deleted all  of the `'a`'s from the signatures, I was in a rush to get it to compile, so I forgot to fix the indentations afterwards.

saghm: I just went through all the files touched by this pull request again, so I think the indentation should be all set.

saghm: ...and of course I missed this one.

PR #89, opened 2015-07-10
Add orderby to find cmd

kyeah:
That awkward moment when you realize the sort option isn't being used.


alabid:
I want to see accompanying tests.

PR #91, opened 2015-07-13
GridFS

kyeah:
Sorry this is all in one PR :(

This is a threaded GridFS implementation on top of the core driver. Chunks are written asynchronously to the database. When a chunk is read from the database, the next chunk in the sequence will be asynchronously loaded and cached.

Will try to provide more detail and tests soon, but opening this for review since there's a lot to look through.


tests/client/gridfs.rs
```diff
@@ -0,0 +1,105 @@
...
+use mongodb::{Client, ThreadedClient};
+use mongodb::coll::options::FindOptions;
+use mongodb::db::ThreadedDatabase;
+use mongodb::gridfs::{Store, ThreadedStore};
+use mongodb::gridfs::file::DEFAULT_CHUNK_SIZE;
+
+use rand::{thread_rng, Rng};
+use std::io::{Read, Write};
+
+#[test]
```
alabid: Awesome test. But I'd like to see some test cases for:
- remove_id/remove
- find

alabid:
Great stuff!!
- I like that you can specify a `chunk_size` or use the default.
- I like your attempt to cache some chunks.
Some options for you to consider:
- Some drivers allow you to specify a different gridfs bucket other than `fs`. But I wouldn't make this a priority since the nodejs driver just uses the default bucket -- https://mongodb.github.io/node-mongodb-native/api-generated/gridstore.html

Just have only one comment about tests.

src/gridfs/file.rs
```diff
@@ -0,0 +1,641 @@
...
+
+        // Complete file write
+        if self.mode  == Mode::Writing {
+            if try!(self.err_description()).is_none() {
+                if self.doc.upload_date.is_none() {
+                    self.doc.upload_date = Some(UTC::now());
+                }
+                self.doc.md5 = self.wsum.result_str();
+                try!(self.gfs.files.insert_one(self.doc.to_bson(), None));
+                //self.gfs.chunks.ensure_index_key("files_id", "n");
```
alabid: why is this commented out? You want to create an index on the "n" and "files_id" fields.

src/gridfs/file.rs
```diff
@@ -0,0 +1,641 @@
...
+                }
+            }
+            cvar.notify_all();
+        });
+
+        Ok(())
+    }
+
+    // Retrieves a binary file chunk from GridFS.
+    fn find_chunk(&mut self, id: oid::ObjectId, chunk: i32) -> Result<Vec<u8>> {
```
alabid: `chunk: i32` should be `chunk_num: i32` or something you deem more appropriate.

src/gridfs/file.rs
```diff
@@ -0,0 +1,641 @@
...
+
+            // Flush chunk to GridFS
+            let mut chunk = self.wbuf.clone();
+            try!(self.insert_chunk(n, &mut chunk));
+            self.wbuf.clear();
+        }
+
+        // Continuously write full chunks of data to GridFS.
+        while data.len() > chunk_size as usize {
+            let size = cmp::min(chunk_size, data.len());
```
alabid: great handling of this gotcha.

src/gridfs/file.rs
kyeah: Oops, sorry. We don't have index methods yet, but I'll add them and make sure they're used in created/checked in GridFS.

alabid: We don't really need to create an index for `n` and `files_id` since we're computing md5 on the client-side. Otherwise, the `filemd5` command on the server requires (or used to) both fields to be indexed.
See: https://github.com/mongodb/mongo/blob/48432412d79d0712367837a12637d3682c04fddb/src/mongo/db/dbcommands.cpp#L582

alabid: Also, see this test that verifies that there's an index on `files_id` and `n` on the `fs.chunks` collection.
https://github.com/mongodb/mongo/blob/master/jstests/core/filemd5.js

kyeah: Alright, so I don't need to ensure the index keys, then? I assume any other GridFS tool that needs them will create these indexes if they don't already exist on the collections. 

I'm guessing it would still be ideal to have these indexes to speed up the fs.

alabid: Yep, please create these indexes to speed up iterating over chunks. You can create an issue for creating indexes and defer its implementation if you want.

src/gridfs/file.rs
```diff
@@ -0,0 +1,641 @@
...
+        };
+
+        // Insert chunk asynchronously into the database.
+        let arc_gfs = self.gfs.clone();
+        let arc_mutex = self.mutex.clone();
+        let arc_wpending = self.wpending.clone();
+        let cvar = self.condvar.clone();
+        let err = self.err.clone();
+
+        thread::spawn(move || {
```
vkarpov15: I'm not entirely sure we want to make one part of the driver async and the rest of the driver sync. What's your reasoning for making this async?

kyeah: The GridFS implementation runs specifically through the Read/Write traits defined by rust, which requires `write()` and `flush()` methods. The `flush()` method in particular is used for buffered writers such as GridFS files, which are written in chunks and don't write the entire file to the database until it's closed/flushed, so it makes sense to make `write` async, provide the use of `flush` if the user needs to block and ensure all bytes have been written, and `close` if they need the entire file to be stable within GridFS.

src/gridfs/file.rs
```diff
@@ -0,0 +1,640 @@
...
+use std::sync::atomic::{AtomicIsize, ATOMIC_ISIZE_INIT, Ordering};
+
+pub const DEFAULT_CHUNK_SIZE: i32 = 255 * 1024;
+pub const MEGABYTE: usize = 1024 * 1024;
+
+/// File modes.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Mode {
+    Closed,
+    Reading,
```
alabid: Why not `Read`, `Write` and not the gerund forms.

src/gridfs/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+pub trait ThreadedStore {
+    /// A new GridFS store within the database with prefix 'fs'.
+    fn with_db(db: Database) -> Store;
+    /// A new GridFS store within the database with a specified prefix.
+    fn with_prefix(db: Database, prefix: String) -> Store;
+    /// Creates a new file.
+    fn create(&self, name: String) -> Result<File>;
+    /// Opens a file by filename.
+    fn open(&self, name: String) -> Result<File>;
+    /// Opens a file by object ID.
```
alabid: nit: s/object ID/ObjectId

src/gridfs/file.rs
```diff
@@ -0,0 +1,640 @@
...
+            self.wsum.input(buf);
+
+            // Pending megabyte
+            while self.doc.chunk_size * self.wpending.load(Ordering::SeqCst) as i32 >= MEGABYTE as i32 {
+                guard = match self.condvar.wait(guard) {
+                    Ok(guard) => guard,
+                    Err(_) => return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        PoisonLockError
+                            ))
```
alabid: indentation here looks weird.

src/gridfs/file.rs
```diff
@@ -0,0 +1,640 @@
...
+            let chunk_num = self.chunk_num;
+            self.chunk_num += 1;
+            self.wsum.input(buf);
+
+            // If over a megabyte is being written at once, wait for the load to reduce.
+            while self.doc.chunk_size * self.wpending.load(Ordering::SeqCst) as i32 >= MEGABYTE as i32 {
+                guard = match self.condvar.wait(guard) {
+                    Ok(guard) => guard,
+                    Err(_) => return Err(io::Error::new(
+                        io::ErrorKind::Other, ArgumentError("test".to_owned())))
```
alabid: "test".to_owned()?

PR #93, opened 2015-07-15
Index operations & refactor tests to drop collections

kyeah:
Implements all create and drop index functions + `list_indexes()`. GridFS now uses `create_index` to speed up iteration over chunks.

Note that once the BSON Encoder/Decoder is built into bson-rs, `IndexModel.to_bson()` can be replaced by a one-liner, `bson::encode(model)`.

This PR also switches the tests to drop their collections instead of the entire database, and fixes the persistence of `wire_protocol` tests by using `drop_db`.

### DIFF
--- a/src/gridfs/file.rs
+++ b/src/gridfs/file.rs
@@ -9,6 +9,7 @@ use Error::{self, ArgumentError, OperationError, PoisonLockError};
 use Result;
 
 use super::Store;
+use coll::options::IndexOptions;
 
 use std::{cmp, io, thread};
 use std::error::Error as ErrorTrait;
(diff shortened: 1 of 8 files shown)


src/gridfs/file.rs
```diff
@@ -192,6 +192,7 @@ impl File {
                 }
                 self.doc.md5 = self.wsum.result_str();
                 try!(self.gfs.files.insert_one(self.doc.to_bson(), None));
+                try!(self.gfs.chunks.create_index(doc!{ "files_id" => 1, "n" => 1}, None));
```
alabid: Create a unique index: https://github.com/mongodb/mongo/blob/v2.6/src/mongo/client/gridfs.cpp#L62

alabid: Context: the  mongo tools (like mongofiles, mongoimport) in versions before 3.0 were built with the mongo binaries. But >= v3.0 has "relegated" the tools (including their accompanying files like gridfs.{h/cpp} to the mongo tools) repo.

alabid: Also create an index on BSON( "filename" << 1 )

PR #94, opened 2015-07-15
Return specific cursor errors

kyeah:
If a cursor command fails, it should return specific operational errors instead of the default CursorNotFoundError.


src/cursor.rs
```diff
@@ -57,6 +57,18 @@ impl Cursor {
                                documents: docs } => {
                 let mut v = VecDeque::new();
 
+                if !docs.is_empty() {
+                    if let Some(&Bson::I32(ref code)) = docs[0].get("code") {                        
+                        // If command doesn't exist or namespace not found, return
+                        // an empty array instead of throwing an error.
+                        if *code == 59 || *code == 26 {
```
alabid: I don't like the way we are comparing `code` to literal numbers. Within the server, most error codes are defined here: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err
so `ErrorCodes::CommandNotFound == 59` and `ErrorCodes::NamespaceNotFound == 26`

alabid: Sample use: https://github.com/mongodb/mongo/blob/07efce25ec43bcd13d565aefad926f8537ab789c/src/mongo/client/dbclient.cpp#L396

kyeah:
Added the ErrorCode type in b3527c1; calling `as i32` is required since the `FromPrimitive` and `ToPrimitive` auto-conversion traits are unstable.

PR #95, opened 2015-07-27
Made tests/client/db.rs actually used

saghm:
Apparently `tests/client/db.rs` was never actually included in the test suite because it wasn't declared as a module in `tests/client/mod.rs`. This pull request fixes that (and a slight issue in the test that occured due to an API change since the test file was written).


alabid:
LGTM. Merge at will.

PR #96, opened 2015-07-27
Changed it so first test in set isn't dropped

saghm:
(I'm not proud of this one...)

Apparently for the entire time since it was first merged, the CRUD spec test suite has been dropping the first test from each set. As far as I can tell, I had it do that to run a certain test, and then forgot to take it out. Luckily, everything still passes.


alabid:
Oh man. Good catch.
LGTM.

PR #99, opened 2015-07-28
APM spec

saghm:


saghm:
@vkarpov15 I added the duration measurement and a simple test for it, if you want to take a look

src/apm/listener.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn add_completion_hook(&self, hook: fn(&CommandResult)) -> Result<()> {
+        let mut guard = match self.completion_hooks.write() {
+            Ok(guard) => guard,
+            Err(_) => return Err(Error::PoisonLockError)
+        };
+
+        Ok(guard.deref_mut().push(hook))
+    }
+
+    pub fn run_start_hooks(&self, started: &CommandStarted) -> Result<()> {
```
vkarpov15: I'd add a boolean flag to check if there are any start hooks first, to avoid having to touch the mutex and potentially getting an error if you haven't set any hooks.

src/error.rs
```diff
@@ -126,6 +133,13 @@ impl fmt::Display for Error {
             &Error::ResponseError(ref inner) => inner.fmt(fmt),
             &Error::CursorNotFoundError => write!(fmt, "No cursor found for cursor operation."),
             &Error::PoisonLockError => write!(fmt, "Socket lock poisoned while attempting to access."),
+            &Error::CodedError(ref err) => write!(fmt, "{}", err),
+            &Error::EventListenerError(ref err) => match err {
+                &Some(ref e) => write!(fmt, "Due to a poisoned lock on the \
```
vkarpov15: nittiest of the nitpicks: "Unable to emit failure due to poisoned lock" is more concise. Commas in error messages read poorly.

src/apm/listener.rs
saghm: Do you mean add another field to the actual client struct? There isn't any way to get the length of the array without the mutex.

tests/apm/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    let db = client.db("test");
+    let coll = db.collection("event_hooks");
+    coll.drop().unwrap();
+
+    let docs = (1..4).map(|i| doc! { "_id" => i, "x" => (rand::random::<u8>() as u32) }).collect();
+    coll.insert_many(docs, false, None).unwrap();
+    client.add_completion_hook(three_second_query).unwrap();
+
+    let doc = doc! {
+        "$where" => (Bson::JavaScriptCode("function() { sleep(1000); }".to_owned()))
```
vkarpov15: I'd recommend we chop this down to say 50ms - no need for the full 1 second sleep and that'll make tests way slower than necessary.

src/apm/listener.rs
vkarpov15: I'm suggesting adding 2 booleans, say `has_start_hook` and `has_completion_hook`, that get set when a start hook / completion hook is added, and check those booleans before trying to acquire the lock. That way we minimize overhead and make it impossible to get a poisoned lock if you haven't set any hooks.

tests/apm/mod.rs
saghm: Given that the tests were taking around 30-40 ms by default, I don't think that length is long enough to accurately test.

src/error.rs
saghm: The reason I had done it like this was because it seemed like the only way to make the error message consist between having a wrapped error and not.

src/apm/listener.rs
saghm: Right, I'm just asking where you intend for those booleans to actually be placed.

tests/apm/mod.rs
vkarpov15: Ok, we'll keep it this way and change it later if it gets to be too much of a pain.

src/apm/listener.rs
vkarpov15: In the struct seems like a logical place.

saghm: Hmmm...looks like this might actually not be feasible. Since `Client` is just an `Arc`, it's fields are enforced as immutable, hence the need for the mutex on the arrays in the first place. In order to change the boolean, it would have to *also* be wrapped in a mutex, which kind of defeats the purpose :/

kyeah: [AtomicBool](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html) may be of use here.

saghm:
Assuming travis passes (which hopefully it should, since the tests passed locally), I think this branch is ready to merge, assuming there are no more code issues.

PR #113, opened 2015-08-11
Add ClientOptions and fix discrepency in server selection timeout

kyeah:
Converted client constructors to use a separate ClientOptions struct. It was already becoming unwieldy with read_preference, write_concern, and log_file, but allowing users to specify SDAM and server-selection options made this a welcome change. The four constructors are now:

```rust
Client::connect(host, port)
Client::connect_with_options(host, port, options)
Client::with_uri(uri)
Client::with_uri_and_options(uri, options)
```

all of which still run through `Client::with_config`.

Also fixed an issue where server selection would return immediately if it failed while retrieving a server for a write operation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,8 @@ use connstring::ConnectionString;
 use db::{Database, ThreadedDatabase};
 use error::Error::ResponseError;
 use pool::PooledStream;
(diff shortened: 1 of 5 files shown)


saghm:
@vkarpov15 I'd like to get this merged soon if possible so I can use the changes in the proof-of-concept tool that I'm writing.

PR #114, opened 2015-08-12
Fix type mismatch on is_master result

kyeah:
Fixes a monitoring bug found by @saghm while building the import/export tool -- the server and topology information was not being updated, because the IsMasterResult would return `ok` as a FloatingPoint instead of an I64 like the automated tests.


kyeah:
Merging; super small fix.

PR #115, opened 2015-08-12
Organized documentation and added module-level documentation

kyeah:
* Added documentation for most modules (inner modules are mostly single-line descriptions).
* Added documentation for some structs and methods that were missing it.
* Organized documentation, e.g.
    * Moved Client and Database function documentation to the traits, so that they are actually shown in the docs.
    * Moved argument documentation from wire_protocol functions to the actual enum fields for OpQuery, OpUpdate, etc.

You can check the documentation [here](http://kyeah.github.io/mongo-rust-driver-prototype/mongodb/) on my personal gh-pages branch. Not sure if we want to preserve the old gh-pages as a separate branch or just make a commit on top of it?


kyeah:
Fixed documentation tests :+1:

PR #116, opened 2015-08-12
Use create option for Collection::new

kyeah:
We should use this option instead of lying in the documentation/comment :)

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -35,12 +35,17 @@ impl Collection {
     /// Creates a collection representation with optional read and write controls.
     ///
     /// If `create` is specified, the collection will be explicitly created in the database.
-    pub fn new(db: Database, name: &str, _create: bool,
+    pub fn new(db: Database, name: &str, create: bool,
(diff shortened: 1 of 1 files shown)


src/coll/mod.rs
```diff
@@ -35,13 +35,18 @@ impl Collection {
...
+    pub fn new(db: Database, name: &str, create: bool,
                read_preference: Option<ReadPreference>,
                write_concern: Option<WriteConcern>) -> Collection {
 
         let rp = read_preference.unwrap_or(db.read_preference.to_owned());
         let wc = write_concern.unwrap_or(db.write_concern.to_owned());
 
+        if create {
+            // Attempt to create the collection explicitly, or fail silently.
+            let _ = db.create_collection(name, None);
```
saghm: Suggestion: instead of ignoring the result (and potentially failing to report an error), maybe there should be a separate method that actually returns a result for when you want to create the collection? It seems to me that if someone feels strongly enough to actually specify the collection should be created, they'll want to know if it failed.

PR #126, opened 2015-10-17
Formatting with Clippy

kyeah:
Ran the build through [rust-clippy](https://github.com/Manishearth/rust-clippy) and fixed up everything that was caught and could be fixed. Mostly involved dereferencing top-level match values (`match *self`) to avoid prefixing `&` to every match case, but also includes:

* Creating StartHook and CompletionHook function type aliases
* Replacing `let ref x = y` with `let x = &y`
* Replacing `for x in y.iter()` with `for x in &y`
* Returning implicitly where possible
* Using implicit lifetimes where possible
* Adding `is_empty()` methods for structs that have `len()`


vkarpov15:
Sweet, thanks Kevin :)

PR #127, opened 2015-10-29
Some changes to reflect Rust 1.4 update

saghm:
Version 1.4 of Rust just came out today, and there are a few changes that affect the driver. First, the use of `Self` as a type in a trait now produces a warning if the trait does not also implement `Sized`. Additionally, the `.expect()` method (essentially `unwrap()` with a custom error message) has been added to `Result`, so there's no need to convert them to `Option` using `.ok()` beforehand.

Since the change with `.expect()` would not work on earlier versions of the compiler, it might make sense to put these changes in a separate `1.4` branch rather than merging with 1.0, which would make them available for people who opt to use newer version of Rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,35 +1,39 @@
-[![Travis](https://travis-ci.org/mongodb-labs/mongo-rust-driver-prototype.svg)](https://travis-ci.org/mongodb-labs/mongo-rust-driver-prototype)
-[![Crates.io](https://img.shields.io/crates/v/mongodb.svg)](https://crates.io/crates/mongodb)
-[![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](LICENSE)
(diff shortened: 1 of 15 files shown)


vkarpov15:
@saghm not quite sure how this would work with releasing this to crates - we can only have one "v0.1.1" on crates as far as I know. It might be worthwhile to merge this into master and just remove support for rust <= 1.3, since there isn't much reason to not upgrade to 1.4 and these changes are just in tests. What do you think?

saghm:
Sounds good, although I'd suggest that we release it as 0.2.0 rather than 0.1.1. Cargo will by default use the highest release available (i.e. if you ask for 0.1.0 and 0.1.1 is available, it'll give you 0.1.1), so we might as well preserve the option of using 0.1.0 with older versions of Rust.

vkarpov15:
As far as I understand it, the driver with this PR will still work on Rust 1.0, just the tests won't work. Is that correct?

saghm:
Yeah, the only changes made in the `src/` directory are the documentation tests, so none of the actual driver implementation is changed.

PR #128, opened 2015-11-02
Fixes for RFC 1214

DenisKolodin:
This PR updates dependencies and some traits to prevent future hard error of RFC 1214.
chrono 0.2.15 - also used to have `Self` with unsized traits, but in 0.2.16 that had fixed.


vkarpov15:
Thanks :beers:

PR #131, opened 2015-12-12
Updated for Mongo 3.2

saghm:
Although around half of the tests failed when I initially ran the tests while connected to  3.2 server, it turns out there were only two small issues.

The first issue was that all errors seem to include error codes in 3.2. Because this wasn't the case in 3.0, there was some code in `cursor.rs` that caused responses with error codes not to return any documents;  this caused some issues with ignoring errors for dropping non-existent collections and attempting to create a collection that already exist. Fixing this fixed almost all of the broken tests, since any tests that dealt with the actual data in collections made sure to drop the collection before putting in the the requisite data for the test.

The remaining two tests were `create_collection` and `list_collections`, which failed because `system.indexes` is no longer returned as one of the collections. Although this is easy to fix in the tests, it raised the issue of needing to make separate test cases for those two depending on the versions. Although no other test cases failed from changes between 3.0 and 3.2, it's always possible that future changes to the driver could cause errors in one but not the other, so I opted to duplicate all of the test cases for 3.2. I also updated the Travis configuration to run both suites, each with the respective server version.


kyeah:
You should be able to use [config options](https://doc.rust-lang.org/book/conditional-compilation.html) to specify 3.2-only tests, e.g. 

```rust
#[test]
fn create_collection() { 
    ...
    if cfg!(not(feature = "mongodb_3_2")) {
        // Check system.indexes
    }
    ...
}

cargo test --features "mongodb_3_2"
```

vkarpov15:
Great work Sam, thanks.

Issue #146, opened 2016-04-29
MapReduce support ?


zerosign:
Is this crate does support mongodb map reduce support or only support Aggregation ? Since in https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/d36d71a2e82f65cf3e2344e4ffa71ce85d4ec0ac/src/command_type.rs, I don't see command type for map reduce in mongodb.

CMIIW

saghm:
You're correct, we did not implement map reduce in the driver. From looking at the [documentation](https://docs.mongodb.org/manual/reference/command/mapReduce/), it seems like it would only require adding the new command type and a corresponding method to the `Collection` type. I'm going to have a lot of free time relatively soon, so if this would be useful for you, I might be able to take a stab at it within the next week or so. Alternately, if you happen to have the time and inclination, a PR would definitely be welcome as well!

zerosign:
I'll try my best this weekend looking the docs and try implement it :+1:

saghm:
I just opened a PR with an implementation of the map-reduce functionality. Feel free to give some feedback about the API defined in it (i.e. the methods defined in `coll/mod.rs`), since you'll probably be the first one to use it once it gets merged!

zerosign:
Sorry, lately, I need to deal with something elses, I'll try to look the PR, thank you.

saghm:
Given that aggregation is strongly preferred over the map reduce feature nowadays (and that it doesn't seem like there's a lot of demand for this feature), I'm going to close this issue.

PR #147, opened 2016-05-06
Implementation of map-reduce functionality

saghm:
As per the request in #146, I've implemented functionality for the [map-reduce command](https://docs.mongodb.org/manual/reference/command/mapReduce/) in the driver. It functions similarly to the other commands, except for the funkiness with the "out" option. Basically, the command can take either a string or a document; in the former case, it dumps the output into the collection specified in the string, and in the latter, the document can specify more options, like dumping to a different database or specifying how to merge/overwrite the existing documents in the collection. I'm not completely convinced there isn't a better design for the signatures of the collection methods and the options struct implemented here, so feedback on that would be appreciated.

Pinging @kyeah as usual


src/coll/options.rs
```diff
@@ -163,6 +163,58 @@ pub struct UpdateOptions {
...
+#[derive(Clone)]
+pub enum MapReduceOutOption {
+    Action(bson::Document),
+    Collection(String),
+}
+
+// Options for map-reduce queries.
+#[derive(Clone)]
+pub struct MapReduceOptions {
+    pub finalize: Option<Bson>,
```
zerosign: Could we just use `Option<Bson::JavascriptCode>` in finalize here? Since, the [specs](https://docs.mongodb.com/manual/reference/command/mapReduce/#mapreduce-finalize-cmd) said so.

src/coll/mod.rs
```diff
@@ -297,6 +297,70 @@ impl Collection {
...
+    }
+
+    /// Runs a "map-reduce" query on the collection.
+    ///
+    /// # Arguments
+    ///
+    /// `map` - A JavaScript function that selects the data to be transformed.
+    /// `reduce` - A JavaScript function that transforms the selected data
+    /// `options` - Other options for the map-reduce query.
+    pub fn map_reduce_with_options(&self, map: &str, reduce: &str, options: MapReduceOptions) -> Result<bson::Document> {
```
zerosign: when dealing with `{ "out" : { "inline" : 1} }`, some folks may prefer to have a `Result<Cursor>` if the document are still big enough to be fetched as `bson::Document`, just like aggregation did.

src/coll/options.rs
```diff
@@ -163,6 +163,58 @@ pub struct UpdateOptions {
 
 pub type ReplaceOptions = UpdateOptions;
 
+// Option for map-reduce "out" argument.
+#[derive(Clone)]
+pub enum MapReduceOutOption {
+    Action(bson::Document),
```
zerosign: We may want to add `Inline` option as enumeration here.

CMIIW

vkarpov15:
lgtm from my end :+1:

zerosign:
@saghm I've a rough sketch for implementation that try to differentiate some arguments that we have when dealing with mongodb map reduce. You might want to check it :D 

https://github.com/mongodb-labs/mongo-rust-driver-prototype/compare/1.0...zerosign:feature/map-reduce?expand=1

src/coll/options.rs
saghm: That's a really good idea; it definitely would be much cleaner that way

saghm:
I like a lot of the stuff in there! Taking the query as an argument probably makes sense, since I'd imagine it's pretty commonly used. I also like the changes with `MapReduceOutput`. Making a separate `Inline` variant definitely makes things a bit more usable.

As for the `MapReduceFn` struct, I'm still learning toward having `map` and `reduce` be separate arguments; it's structured that way in the Mongo shell, and it's also more consistent with how the other `Collection` methods are implemented. As for the `finalize` function, I'm not sure if that should be an argument or an option. How common is passing a `finalize` function in making map reduce queries in Mongo (I haven't really used them much). If they're fairly uncommon, I'd lean towards making them an option rather than an argument, but I don't really have a good enough sense of the common forms of the map reduce queries to know which way this would go.

I think the best way to go forward might be for me to merge my changes into a separate branch in this repo and then for you to make a PR to it so we can iron out the differences between the two implementations. Do you think you'll have some spare time any time soon to work on this a bit more?

kyeah:
Agreed with Sam -- we should be sticking close to the mapReduce implementations of other drivers, namely, taking in separate `map` and `reduce` arguments. We should also try to make the API as uniform as possible, and I think making the query and lesser-used options separate structs moves away from that. Perhaps keeping `sort` and `limit` in the options, and taking in the query explicitly as an argument may make more sense in keeping it close to the `find` implementation?

zerosign:
@saghm I'll try :-)
@kyeah ok

kyeah:
hey @saghm @zerosign ! It's been a while :) Just wondering if there's any potential or interest in picking this back up. If not, I'm happy to play around with different implementations and see what gels.

saghm:
(@kyeah Honestly, I completely forgot this was still pending. My bad)

@zerosign, *wayyyyy* back, I had suggested that you make a PR merging my changes and yours. I think I was a little unclear about it though; I meant for the PR to be made to this repo, not my fork. If you're still interested in helping us out with that, feel free to submit the PR here instead! Otherwise, I can take care of it

zerosign:
@kyeah @saghm sorry for not following this a bit, sorry to not be able to contribute to this project at this time :'(. I think I'm okay with anyone to take whether to merge this PR or merge my implementation with this PR.

PR #149, opened 2016-05-13
Forward rust-nightly feature to textnonce

Gyscos:
This allows to use `serde_macros` without conflict from the `serde_codegen` dependency that would otherwise be present in `textnonce`.
To compile on nightly :

```
cargo build --no-default-features --features rust-nightly
```

Or in `Cargo.toml`:

```toml
[dependencies]
mongodb = { version = "0.1.4", default_features = false, features = ["rust-nightly"] }
```


kyeah:
LGTM. Thanks @Gyscos! :+1:

PR #154, opened 2016-08-10
Use `setVersion` to detect stale topology updates

### LINKED ISSUE #153: SDAM test failing after update to submodule
After updating the `specifications` submodule, I noticed that one of the SDAM tests is failing. Specifically, the last phase of `rs/new_primary_new_setversion.json` is expecting a different set name for the server. From what I can tell, the set name given in the outcome for the server with hostname "a:27017" is not the same as it is in the response.

@kyeah, any idea if the error is due to a mistaken assumption on our part in terms of the test? I read through `https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests`, but I'm still not completely sure if something has changed in terms of the server responses or if our tests are just not handling this case properly.

saghm:
I added a `set_version` field to `Server` and `IsMaster` result as well as a `max_set_version` field to `TopologyDescription` to be used in the same way as `election_id` and `max_election_id` to detect stale updates to the topology. This fixes #153.

The SDAM spec tests still fail after this due to a later test in the suite failing; I double checked that this update didn't cause the later failure by temporarily removing the test this PR fixes and running the test suite against HEAD, and the later test still fails (I'll create an issue about it once I have a chance to look into it).

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -57,6 +57,8 @@ pub struct TopologyDescription {
     pub server_selection_timeout_ms: i64,
     // The largest election id seen from a server in the topology.
     max_election_id: Option<oid::ObjectId>,
+    // The largest set version seen from a server in the topology.
+    max_set_version: Option<i64>,
     // If true, all servers in the topology fall within the compatible
     // mongodb version for this driver.
     compatible: bool,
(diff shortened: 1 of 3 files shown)


saghm:
Actually, this is inconsistent with the spec. I'll make a new PR with the correct changes.

PR #155, opened 2016-08-10
Hidden replica set members should be RsOther

saghm:
According to [the spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#type), hidden replica set members should be of type RSOther. [One of the tests](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/rs/rsother_discovered.json) had been updated to test this.


kyeah:
LGTM :ok_hand:

PR #156, opened 2016-08-10
Fix update rs with primary from member

saghm:
According to [the spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#user-content-updaterswithprimaryfrommember), the return statement should be in the second if, not the first. This fixes the failure in the test `stepdown_change_set_name.json`.


kyeah:
Also LGTM :ok_hand:

PR #157, opened 2016-08-10
Implement using `setVersion` to detect stale primaries

### LINKED ISSUE #153: SDAM test failing after update to submodule

saghm:
Updates monitoring to use `set_version` as per [the spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary) to fix #153. Once this and #155 and #156 are merged, all the SDAM tests should pass.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -60,6 +60,8 @@ pub struct TopologyDescription {
     // If true, all servers in the topology fall within the compatible
     // mongodb version for this driver.
     compatible: bool,
+    // The largest set version seen from a primary in the topology.
+    max_set_version: Option<i64>,
     compat_error: String,
 }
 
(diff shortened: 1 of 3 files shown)


kyeah:
Oof, that if block. Welp, if that's what they provide in the spec, LGTM :)

saghm:
Yeah, I was trying to decide if it was worth splitting it into another nested if statement, but I thought I should try to make it as close to the spec as possible

PR #158, opened 2016-08-11
Fix GridFS race condition

saghm:
The GridFS cache [get initialized with an empty chunk](https://github.com/saghm/mongo-rust-driver-prototype/blob/fix-gridfs-race-condition/src/gridfs/file.rs#L338) but the chunk itself is initialized asynchronously, which leads to a race condition when the cache is checked [here](https://github.com/saghm/mongo-rust-driver-prototype/blob/fix-gridfs-race-condition/src/gridfs/file.rs#L292) on subsequent calls to `get_chunk`. This is fixed by initializing each `CachedChunk` with a placeholder error that gets removed when the chunk is initialized.

### DIFF
--- a/src/gridfs/file.rs
+++ b/src/gridfs/file.rs
@@ -649,7 +649,7 @@ impl CachedChunk {
         CachedChunk {
             n: n,
             data: Vec::new(),
-            err: Some(Error::DefaultError(String::from("Chunk has not yet been inititalied"))),
+            err: Some(Error::DefaultError(String::from("Chunk has not yet been initialized"))),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


vkarpov15:
Lgtm. In case you're waiting for me to review, just feel free to merge this and future prs on your own :)

src/gridfs/file.rs
```diff
@@ -646,7 +649,7 @@ impl CachedChunk {
         CachedChunk {
             n: n,
             data: Vec::new(),
-            err: None,
+            err: Some(Error::DefaultError(String::from("Chunk has not yet been inititalied"))),
```
kyeah: typo!!! `inititalied`

saghm: Ahhh! I'm so embarrassed

saghm:
@kyeah, everything else look good? All the tests pass now that the other PR's have been merged

kyeah:
yeah, go for it!

saghm:
I'm thinking we should push a new version to crates?

kyeah:
yup, looks like a good time for push out 0.1.5.

PR #161, opened 2016-09-02
fix topology description race condition

kali:
Hi,

I came upon a race condition in the topology monitoring. It manifests itself while initializing the driver against a replica set (not with a "fake" one on localhost, I could only reproduce it with a more real setup with a bit of lan latency).

What happen basically is that the self.description is locked while we make the current thread sleep, so we need to be very lucky to get it to get updated. Moving the first part of the loop in an ad-hoc scope, confining the lock there, and putting the sleep outside of this scope solved the issue for us. 

The diff looks bad but most of it is reindenting the main logic inside its ad-hoc scope.

Tell me what you think.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -731,29 +731,31 @@ impl Topology {
         let start_ms = time.sec * 1000 + (time.nsec as i64) / 1000000;
 
         loop {
-            let description = try!(self.description.read());
-            let result = if write {
-                match description.acquire_write_stream() {
-                    Ok(stream) => Ok((stream, false, false)),
-                    Err(err) => Err(err),
-                }
-            } else {
-                description.acquire_stream(read_preference.as_ref().unwrap())
-            };
-
-            match result {
-                Ok(stream) => return Ok(stream),
-                Err(err) => {
-                    // Check duration of current server selection and return an error if overdue.
-                    let end_time = time::get_time();
-                    let end_ms = end_time.sec * 1000 + (end_time.nsec as i64) / 1000000;
-                    if end_ms - start_ms >= description.server_selection_timeout_ms {
-                        return Err(err)
+            {
+                let description = try!(self.description.read());
+                let result = if write {
+                    match description.acquire_write_stream() {
+                        Ok(stream) => Ok((stream, false, false)),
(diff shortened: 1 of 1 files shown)


saghm:
Oh, wow, you're right, Github definitely is having trouble presenting this diff in a readable way.

This intuitively seems like it would fix the problem, but I'd be interested in trying it out myself to verify this. How exactly did you test it? Would making a simple three-node set and then shutting down one of them cause the issue (assuming the latency is sufficient)?

kali:
Nothing fancy on my side. The replica set is made of three instances on google compute cloud, each on a different zone (it may incur a tiny bit of latency), and the client is on a fourth machine. The cheapest ones. I did not need to take one down to see the problem, and it was showing let's say... 95% of the time. A three replica setup on my laptop did not make it happen.

Note that while debugging, a simple println! at the very beginning of the loop was enough to "fix" the problem. After one or two iterations, the description was updated from unknown to ReplicaSetWithPrimary.

kali:
(The fourth machine was a google instance too)

kyeah:
hey! just noticed this long-standing PR. @saghm, have you been able to replicate kali's scenario in any format?

Regardless, not sleeping with a held lock seems like something we should do regardless in this situation. thoughts on merging this?

saghm:
@kyeah Ugh, I keep meaning to try and replicate this, but I haven't had a chance to yet.

Do you think we need to replicate it before merging? Given that this was code I hadn't really touched before, I was leaning towards replicating it first before merging, but if you think it looks fine, I'd be all right with merging it

kyeah:
I'm comfortable merging. Sorry for the wait, @kali!

kali:
thanks! and no worries for the delay, cargo github support makes this
relatively painless.


On Wed, Oct 12, 2016, at 05:42, Kevin Yeh wrote:
> Merged #161[1].
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/mongodb-labs/mongo-rust-driver-prototype/pull/161
  2. https://github.com/mongodb-labs/mongo-rust-driver-prototype/pull/161#event-820435191

PR #168, opened 2016-10-15
Format and lint

saghm:
I used `rustfmt` and `clippy` to clean up the codebase. Obviously this is a ton of changes, and it's probably not worth going through all of them. All of the tests passed locally though, so unless @kyeah has any objections, I'll merge this if Travis reports everything as okay


kyeah:
rustfmt / clippy have been a little iffy in the past. i have hopes that most of the common issues are fixed, but I'm going to make a quick pass at the formatting once i get to my computer.

saghm:
Sounds good. FYI there were some things reported that I didn't fix (e.g. a few lines that were reported as too long that couldn't be shortened easily and functions that had "too many arguments")

src/auth.rs
```diff
@@ -207,25 +225,29 @@ impl Authenticator {
...
+                        return Err(ResponseError("Invalid UTF-8 payload returned".to_owned()))
+                    }
                 };
 
                 // Check that the signature exists
                 let verifier = match scan_fmt!(&payload_str[..], "v={}", String) {
                     Some(string) => string,
-                    None => return Err(MaliciousServerError(MaliciousServerErrorType::NoServerSignature)),
+                    None => return Err(
+                        MaliciousServerError(MaliciousServerErrorType::NoServerSignature)),
```
kyeah: ugh, this doesn't feel consistent with the change above where it puts it in a block/scope, but I know it's a line length thing. :ok:

src/coll/error.rs
```diff
@@ -320,17 +332,19 @@ impl BulkWriteException {
         // Parse out any write errors.
         let w_errs = if let Some(&Bson::Array(ref errors)) = result.get("writeErrors") {
             if errors.is_empty() {
-                return Err(Error::ResponseError(
-                    "Server indicates a write error, but none were found.".to_owned()));
+                return Err(Error::ResponseError("Server indicates a write error, but none were \
+                                                 found."
+                    .to_owned()));
```
kyeah: actually weird formatting here, especially since the original seemed to be within the character limit.

src/coll/error.rs
```diff
@@ -320,17 +332,19 @@ impl BulkWriteException {
...
             for err in errors {
-                if let &Bson::Document(ref doc) = err {
+                if let Bson::Document(ref doc) = *err {
                     vec.push(try!(BulkWriteError::parse(doc.clone())));
                 } else {
-                    return Err(Error::ResponseError(
-                        "WriteError provided was not a bson document.".to_owned()));
+                    return Err(Error::ResponseError("WriteError provided was not a bson \
+                                                     document."
+                        .to_owned()));
```
kyeah: same as above -- weird formatting.

src/coll/mod.rs
```diff
@@ -66,9 +69,15 @@ impl Collection {
...
-                                        .nth(idx+1).unwrap().0..].to_owned(),
+        match self.namespace.find('.') {
+            Some(idx) => {
+                self.namespace[self.namespace
+                        .char_indices()
+                        .nth(idx + 1)
+                        .unwrap()
+                        .0..]
+                    .to_owned()
+            }
```
kyeah: this seems like it's harder to read? :/

src/coll/mod.rs
```diff
@@ -335,108 +425,124 @@ impl Collection {
...
-            Batch::Update(models) =>
-                self.execute_update_batch(models, start_index, ordered, result,
-                                          exception),
+            Batch::Insert(docs) => {
+                self.execute_insert_batch(docs, start_index, ordered, result, exception)
+            }
+            Batch::Delete(models) => self.execute_delete_batch(models, ordered, result, exception),
+            Batch::Update(models) => {
+                self.execute_update_batch(models, start_index, ordered, result, exception)
+            }
```
kyeah: pretty surprised it handled these cases differently -- seems like they're all the same length! Does it just ensure the first and last are in a block if there are more than two cases?

src/cursor.rs
```diff
@@ -100,20 +101,31 @@ impl Cursor {
...
+                      format!("{}.$cmd", db),
+                      1,
+                      OpQueryFlags::no_flags(),
+                      0,
+                      0,
+                      doc,
+                      None,
+                      cmd_type,
+                      true,
+                      read_pref)
```
kyeah: not gonna tackle how ridiculously long our `query` fn is, and how rust doesn't have named args :(

src/auth.rs
saghm: Yeah, the line-length things it does do seem pretty odd. Maybe there's a way to run it with the line-length lint turned off?

src/coll/error.rs
saghm: Yeah, that is quite odd...

From looking at it, I think that should even be changed to use `String::from` (which got stabilized since last summer). Maybe I should do a search for all of the `.to_owned` calls and see which ones are replaceable?

kyeah: this sounds good.

saghm: Just pushed the changes to this branch. A lot of the instances of `.to_owned()` can't be changed due to them being called on `&String` rather than `&str`. Do you think it's worth trying to change those to `.clone()`?

I also decided to hold off on making the changes in the `v3_x` test directories pending the re-merging of them that we discussed offline, since it will take half as much work to do that after they're merged again.

kyeah:
I think we can merge this. There are a couple funky bits and pieces, but the majority of the changes are good and fix existing weirdness in the codebase.

PR #169, opened 2016-11-05
Merge versioned tests

kyeah:
#### 🌴 aloha

Back in #131, we split the tests into `v3_0` and `v3_2` because of a small change in the way `list_collections` returned results: it stopped returning `system.indexes`.

Turns out, having duplicated tests kinda sucks. This PR merges them back together and uses [config features](https://doc.rust-lang.org/beta/book/conditional-compilation.html) to test version 3.0 and 3.2.

#### caveats

Having a config feature _just_ to make testing easy isn't very intuitive for people using the driver -- I can easily see someone thinking they have to use this feature if they're using mongodb 3.0. In reality, the driver will gracefully handle different versions without any user configuration.

I'm going to followup this PR by adding a `db.version` command function which uses [semver](https://github.com/steveklabnik/semver). We can use this in the test to determine the mongodb version at runtime. However, I'd rather do this after merging the tests for simplicity.

#### bson

I've also locked the bson version. I ran into an issue when we automatically updated to bson `0.3.2`. A new `Bson::Symbol` type was added that broke an exhaustive `match` block. I fixed that.


Cargo.toml
```diff
@@ -13,7 +13,7 @@ keywords = ["mongo", "mongodb", "database", "bson", "nosql"]
 license = "Apache-2.0"
 
 [dependencies]
-bson = "0.3"
+bson = "0.3.2"
```
saghm: Looking back, I'm not actually sure why we didn't specify a release for our dependencies. (I'm guessing that I probably thought that specifying the release meant it was pinned rather than that it would let you update releases but not minor versions way back when we first added them).

Given that all our dependencies are < 1.0, I think it would make sense to specify releases for all of them. Want to add in the release numbers for the other three?

kyeah:
Updated release numbers for anything without a patch version 👌.

PR #172, opened 2016-11-19
Use db.version() for test switch

kyeah:
👋 

This removes the test-specific feature flag in favor of using `db.version()`, completing our PR narrative for merging versioned tests. #169 #171

### DIFF
--- a/tests/client/db.rs
+++ b/tests/client/db.rs
@@ -20,11 +20,13 @@ fn create_collection() {
 
     let results = cursor.next_n(5).unwrap();
 
-    let v3_0 = cfg!(feature = "mongodb_3_0");
-    let result_size = if v3_0 { 3 } else { 2 };
+    let db_version = db.version().unwrap();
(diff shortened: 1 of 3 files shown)


tests/client/db.rs
```diff
@@ -20,11 +20,13 @@ fn create_collection() {
 
     let results = cursor.next_n(5).unwrap();
 
-    let v3_0 = cfg!(feature = "mongodb_3_0");
-    let result_size = if v3_0 { 3 } else { 2 };
+    let db_version = db.version().unwrap();
+    let v3_1 = db_version.major <= 3 && db_version.minor <= 1;
```
saghm: Have you tested this on the last 3.1 release? I would have thought that 3.1 would have the same results as 3.2 rather than 3.0

kyeah: yup, tested it on 3.0, 3.1, 3.2.

saghm: LGTM in that case

Issue #164, opened 2016-09-17
Look into making tests parallelizable


saghm:
Right now, the test suites needs to be run with RUST_TEST_THREADS=1. It's probably possible to make them able to run in parallel by doing things like using unique namespaces for each test case, which would be much more friendly.

saghm:
Fixed in #173

PR #173, opened 2016-11-19
Parallelize test suite

kyeah:
🌵:dancer:

## parallelization.

This makes our tests parallelizable by naming databases on a per-module level, and collections on a per-test level. When necessary (mostly in `client/mod` and `client/db`), databases will be named on a per-test level.

## a tale of two servers.

Previously in `client/mod`, we required an extra server on 27018, since we were wiping all databases at the start of each test and checking the number of databases listed. 

I didn't find this very necessary, as long as the databases we _know_ should exist, exist. This includes the base `local`/`admin` databases, along with databases we create ourselves, so it seems to cover the cases for calling `database_names`.

I modified the tests to check these databases specifically, and removed 27018 from the travis config. Less test suite setup 🚀 

## oops.
This includes the db.version() test switch because I forgot to switch to master before starting this branch, and now it's too much work to backtrack and take it out. It's a small change. nbd.

### DIFF
--- a/tests/client/db.rs
+++ b/tests/client/db.rs
@@ -20,11 +20,13 @@ fn create_collection() {
 
     let results = cursor.next_n(5).unwrap();
 
-    let v3_0 = cfg!(feature = "mongodb_3_0");
-    let result_size = if v3_0 { 3 } else { 2 };
+    let db_version = db.version().unwrap();
+    let v3_1 = db_version.major <= 3 && db_version.minor <= 1;
+
+    let result_size = if v3_1 { 3 } else { 2 };
     assert_eq!(result_size, results.len());
 
-    if v3_0 {
+    if v3_1 {
         match results[0].get("name") {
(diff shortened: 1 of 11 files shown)


kyeah:
oh yeah, this fixes #164.

tests/client/db.rs
```diff
@@ -62,32 +67,37 @@ fn list_collections() {
...
 
-    if v3_0 {
+    if v3_1 {
         match results[0].get("name") {
             Some(&Bson::String(ref name)) => assert_eq!("system.indexes", name),
             _ => panic!("Expected BSON string!"),
         };
     }
 
+    let db_1 = if v3_1 { "test" } else { "test2" };
```
saghm: Minor style nitpick (feel free to ignore if you disagree): I think `db1/db2` looks better than `db_1/db_2`

kyeah: ah, we've been using no-underscores. minor slip up here.

PR #175, opened 2016-12-23
Implement SSL

### LINKED ISSUE #167: Add TLS Connections
All,
  As discussed in #165,  the ability to use TLS connections are necessary to use services like MongoDB ATLAS.  I'm not sure if the Tokio-TLS library (https://github.com/tokio-rs/tokio-tls), would help ease the implementation of this, but may be something worth exploring.  Tokio and Tokio-TLS were discussed at the 2016 Rust Conference in this presentation (https://www.youtube.com/watch?v=bcrzfivXpc4&index=4&list=PLE7tQUdRKcybLShxegjn0xyTTDJeYwEkI).  

Thanks!
--Nick

saghm:
Fixes #167.

The most straightforward way to add SSL support to the driver was to use OpenSSL, so that's what I went with. This has the downside of not working on systems without OpenSSL installed, so in the interest of still allowing people on such systems to compile/use the driver, I added SSL as an optional feature. This adds a bit of complexity to the code in the form of needing `[cfg(feature = "ssl")]` in a lot of places, but I think this is preferable to users on systems without SSL installed by default (e.g. Windows) to still use the driver. I also elected to have the driver default to not including SSL support, and an [informative error message](https://github.com/mongodb-labs/mongo-rust-driver-prototype/compare/master...saghm:ssl?expand=1#diff-049b072e95cd0be52f0e7cc2e83a742cR258) is given if the user tries to connect with SSL without enabling it to ensure that nobody thinks that they're using SSL when they're actually not.


src/error.rs
```diff
@@ -71,6 +82,8 @@ pub enum Error {
     /// A standard error with a string description;
     /// a more specific error should generally be used.
     DefaultError(String),
+    #[cfg(feature = "ssl")]
+    SslError(String),
```
saghm: I originally wanted to have something better than just a string for SSL errors, but the openssl crate does not do errors very idiomatically, so returning its error types to users would likely just be confusing for users.

src/pool.rs
```diff
@@ -149,11 +230,37 @@ impl ConnectionPool {
...
         let host_name = &self.host.host_name;
         let port = self.host.port;
-        let stream = BufStream::new(try!(TcpStream::connect((&host_name[..], port))));
-        Ok(stream)
+        let inner_stream = try!(TcpStream::connect((&host_name[..], port)));
+
+        if let Some(SslConfig { ca_file: ref _ca_file,
+                                certificate_file: ref _certificate_file,
+                                key_file: ref _key_file }) = self.ssl {
+            #[cfg(feature = "ssl")]
```
saghm: Putting the feature attribute above a block is very messy, but it seems to be the most ergonomic way to conditionally enable a chunk of code that wouldn't compile if the feature is disabled. I originally intended to use the `if cfg!(...) { ... }` syntax, but it turns out that `cfg!` just expands to either `true` or `false` at compile time, so the code within the block would still be processed by the compiler.

saghm:
It would be doable, although it would be a bit funky. The tests would need need SSL certificates and private keys, so they would either have to be manually added to the repo or generated with a sort of script that would have to be invoked before the test. Then a separate `mongod` would have to be run with the correct arguments (something like `mongod --sslMode requireSSL --sslPEMKeyFile server.pem --sslCAFile ca.pem`) for the client to test against.

If we decide to go this route, our tests would then require 3 separate `mongod`'s, which is starting to get a little unwieldy. It's probably worth looking into using something like [this](https://github.com/10gen/mongo-orchestration) to make setting them up a bit easier.

kyeah:
A set of certs and pkeys confined to an ssl test folder wouldn't be too bad, I think. Perhaps we can do a separate `mongod` for this PR and follow-up with mongo-orchestration?

saghm:
Works for me! I'll push another commit to the PR adding the test, certs, and keys once I get a chance

.travis.yml
```diff
@@ -3,18 +3,29 @@ language: rust
 cache: cargo
 
 before_install:
-    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.0.14.tgz
-    - tar xvf mongodb-linux-x86_64-3.0.14.tgz
-    - mv mongodb-linux-x86_64-3.0.14 3.0.14
-    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.2.11.tgz
-    - tar xvf mongodb-linux-x86_64-3.2.11.tgz
-    - mv mongodb-linux-x86_64-3.2.11 3.2.11
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.0.14.tgz
```
saghm: The old link downloads the generic "Linux legacy" tarball, which doesn't link to openssl, so in order to have Travis test this, we need to use the distro-specific package.

.travis.yml
```diff
@@ -3,18 +3,29 @@ language: rust
...
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.0.14.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.0.14 3.0.14
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.2.11.tgz
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.2.11.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.2.11 3.2.11
 
 script:
-    - mkdir -p ./data/db ./data/db2
-    - 3.0.14/bin/mongod --fork --nopreallocj --dbpath ./data/db --syslog --port 27017
```
saghm: I'm not completely sure why we were doing `--nopreallocj` beforehand; I think back when we first started, I was doing that locally to cut down on the startup time, but this isn't really necessary for Travis, as the 3.0 `mongod`'s are starting up in the background while the code is compiling and 3.2+ uses WiredTiger by default (which doesn't preallocate).

.travis.yml
```diff
@@ -3,18 +3,29 @@ language: rust
...
+    - mkdir -p ./data/db30 ./data/db30-ssl ./data/db32 ./data/db32-ssl
```
saghm: Given that we now need four different databases running for the tests, I thought giving the directories descriptive names would be worthwhile.

src/pool.rs
```diff
@@ -149,11 +230,37 @@ impl ConnectionPool {
         }
     }
 
-    // Connects to a MongoDB server as defined by the initial configuration.
-    fn connect(&self) -> Result<BufStream<TcpStream>> {
+
+    // Connects to a MongoDB server as defined by the initial
```
kyeah: cut off the documentation here. "initial configuration."

src/pool.rs
```diff
@@ -149,11 +230,37 @@ impl ConnectionPool {
...
-        Ok(stream)
+        let inner_stream = try!(TcpStream::connect((&host_name[..], port)));
+
+        if let Some(SslConfig { ca_file: ref _ca_file,
+                                certificate_file: ref _certificate_file,
+                                key_file: ref _key_file }) = self.ssl {
+            #[cfg(feature = "ssl")]
+            {
+                let mut ssl_context = SslContext::builder(SslMethod::tls())?;
+                ssl_context.set_cipher_list("DEFAULT")?;
```
kyeah: should this and sslVerify be optional/defaultable SslConfig params?

src/pool.rs
saghm: Ah, good catch.

src/pool.rs
saghm: From looking at the options that the mongo shell has, the SSL verification options seem to be `--sslAllowInvalidHostnames` and `--sslAllowInvalidCertificates` (which both default to off), whereas the OpenSSL crate has `SSL_VERIFY_NONE`, SSL_VERIFY_PEER`, and `SSL_VERIFY_FAIL_IF_NO_PEER_CERT`. If I correctly understand the OpenSSL crate documetation on these types correctly, then `SSL_VERIFY_PEER` and `SSL_VERIFY_FAIL_IF_NO_PEER_CERT` should be on by default, and `--sslAllowInvalidCertificates` is the equivalent of turning off `SSL_VERIFY_PEER` (but not `SSL_VERIFY_FAIL_IF_NO_PEER_CERT`. I'm having trouble finding the equivalent option to `--sslAllowInvalidHostnames` in the OpenSSL crate, but I'd be all right with leaving out the ability to use this option unless/until someone asks to be able to use it.

The cipher list is a bit trickier; I'm having trouble finding documentation specifying exactly which ciphers are able to be used in MongoDB, but from looking at the options available to `mongo` and `mongod`, I think this is something that's only configurable on the server-side, so this might just be something that drivers don't have a specific API for. I'm going to be back at work tommorrow, so I'll check with some of the drivers people tomorrow to see what the other drivers do with regard to this.

src/pool.rs
kyeah: heyo, re: this, I'm imagining moving this into its own module and having `Stream` be a Trait instead of an enum that's matched on. Something similar to what's being down in [rust-postgres](https://github.com/sfackler/rust-postgres/tree/master/postgres/src/tls). This would move most of the ssl-specific code into a single module.

src/pool.rs
kyeah: any update on the cipher stuff here?

re: hostname verification, I've seen requests on things like reqwest and [rust-native-tls](https://github.com/sfackler/rust-native-tls/issues/13) to disable verification, but I agree that it's an option that can be added in the future if someone requests it. I believe it's set using [bitflags](https://github.com/sfackler/rust-openssl/pull/469/files#diff-a654b9fe1cba8cdeefbbd7d1e925b257R8).

```rust
ssl.param().set_hostflags(X509_CHECK_FLAG_NEVER_CHECK_SUBJECT);
```

re: ciphers, generally clients and servers both have their own accepted cipher list; if they have overlap, the connection is established and any of those overlapping ciphers can be chosen for encryption. These are defined as strings (see [the OpenSSL docs](https://www.openssl.org/docs/man1.0.2/apps/ciphers.html)), so you should be able to accept one in the SslConfig. If the user provides an invalid string, the openssl crate should error out accordingly (i'm guessing).

src/error.rs
kyeah: Do you you have an example of an OpenSSL error in default formatting vs. the one implemented here? Just wondering if it might be preferable to remove the special case and have these errors be coerced to `io::Error` automatically.

kyeah:
:wave: sorry for the delayed follow-up, was waiting to see if there was any followup to ciphers. more thoughts inlined now that I've looked at some other uses of the OpenSSL crate.

src/error.rs
saghm: I'd be okay with that. In their current form, the error messages are admittedly not that helpful.

src/pool.rs
saghm: That makes a lot of sense. I'll rework this and update the PR (along with the other things).

src/pool.rs
saghm: Sorry about the wait! I've talked to some people, and apparently there's a TLS spec in the works. For now, I've gotten some detailed (but not finalized) information about how the drivers are supposed to configure TLS, so I'll set the options the way they're recommended when I update the PR.

There are a number of other things that we'll likely need to implement to be compliant with the spec down the line, so we'll likely have to update a few things once it's finalized. For now though, I think it's better to get a working implementation merged.

kyeah:
@saghm, seeing a lot of branching cases here with the feature flag and `with_ssl` methods. What if we generalized the `ssl: Option<SslConfig>` field + `with_ssl` methods and _required_ a StreamConnector to be passed around? Something like:

```rust
// stream.rs
pub trait StreamConnector {
  fn connect(&self, hostname: &str, port: u16) -> io::Result<Stream>;
}

pub struct TcpStreamConnector;

impl StreamConnector for TcpStreamConnector {
  fn connect(&self, hostname: &str, port: u16) -> io::Result<Stream> {
    TcpStream::connect(host, port);
  }
}

pub struct SslStreamConnector {
  pub ca_file: String,
  pub certificate_file: String,
  pub key_file: String,
  pub verify_peer: bool,
}

impl StreamConnector for SslStreamConnector {
  fn connect(&self, hostname: &str, port: u16) -> io::Result<Stream> {
    // Use ssl file opts
  }
}
```

```rust
// mod.rs
pub struct ClientOptions {
  ...
  pub stream_connector: StreamConnector,
}

impl ClientOptions {
  pub fn new() -> ClientOptions {
    ClientOptions {
      ...
      stream_connector: TcpConnector::new(),
    }
  }

  pub fn with_stream_connector(connector: StreamConnector) -> ClientOptions {
    ...
  }
}

**Comment by saghm** [comment]
Sounds good to me. I can implement that within the next day or so, unless you'd prefer to do it yourself.

**Comment by kyeah** [comment]
np, whenever you have time. sorry about the back-and-forth — wish it was easier to collaborate :)

**Comment by saghm** [comment]
No worries! I was mainly asking if you wanted to make the change because you had basically already written all the necessary code, and I wanted to give you the option of not having me steal your work.

**Comment by saghm** [comment]
So it turns out that using traits for StreamConnector and Stream don't work out that wonderfully. Using StreamConnector efficiently requires that it be clonable, as methods like `TopologyDescription::add_missing_hosts` would need to pass it multiple times. Unfortunately, any trait that implements a method with itself in the type signature (such as `Clone::clone`) or even with a generic type parameter (like `fn clone<T>(&self) -> T where T: StreamConnector`, which is one hacky workaround I tried) aren't able to be used as trait objects. This means that we'd have to pass StreamConnectors around everywhere by generic type parameters with bounds, which in practice turns out to "infect" almost every type in the library (up to and including all the major public-facing ones: Client, Database, Collection, and Cursor). This is pretty suboptimal; requiring users to deal with a generic parameter on Client, Database, Collection, and Cursor for the sake of having cleaner code is questionable, but having users not compiling with SSL doing it for the sake of code that they're not even compiling is just plain rude.

Fortunately, the brilliant idea of passing around a StreamConnector works quite well with the original strategy of using an enum for TCP vs. SSL. The resulting code only requires conditional compilation in two places; `lib.rs`, for `extern crate openssl` and `ClientOptions::with_ssl` and `stream.rs`, for the actual implementation of the SSL variant of the enum.

On a related note, implementing `Stream` as a trait doesn't quite work either; returning a `Result<Box<TcpStream>>` for a `Result<Box<Stream>>` is not allowed, and casting `TcpStream` to `Stream` is only allowed when `Stream` is `Sized`...which brings us back to the original problem of having to use generics instead of trait objects. This is also quite easily fixed by making `Stream` an enum rather than a trait.

**Comment by kyeah** [comment]
ah, awesome! thanks for looking into this — looks like it turned out pretty clean/self-contained overall. I'll do a good pass-through this weekend, if not tonight.

File: src/stream.rs [hunk 8]
```diff
@@ -0,0 +1,127 @@
...
+use std::net::{SocketAddr, TcpStream};
+
+#[cfg(feature = "ssl")]
+use openssl::ssl::{Ssl, SslMethod, SslContext, SslStream, SSL_OP_NO_COMPRESSION, SSL_OP_NO_SSLV2,
+                   SSL_OP_NO_SSLV3, SSL_VERIFY_NONE, SSL_VERIFY_PEER};
+#[cfg(feature = "ssl")]
+use openssl::x509::X509_FILETYPE_PEM;
+
+#[derive(Clone)]
+pub enum StreamConnector {
```
kyeah: docs for the enum and each variant?

src/stream.rs
```diff
@@ -0,0 +1,127 @@
...
+            #[cfg(feature = "ssl")]
+            Stream::Ssl(ref mut s) => s.write(buf),
+        }
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        match *self {
+            Stream::Tcp(ref mut s) => s.flush(),
+            #[cfg(feature = "ssl")]
+            Stream::Ssl(ref mut s) => s.flush(),
```
kyeah: Is it possible to provide a private function that returns this mutable stream ref, so we can just call `self.mut_stream().flush()`? Maybe even impl `DerefMut`?

(theories from someone who's forgotten all about rust intricacies)

src/stream.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+}
+
+
```
kyeah: extra line

src/stream.rs
```diff
@@ -0,0 +1,127 @@
...
+                    -> Self {
+        StreamConnector::Ssl {
+            ca_file: String::from(ca_file),
+            certificate_file: String::from(certificate_file),
+            key_file: String::from(key_file),
+            verify_peer: verify_peer,
+        }
+    }
+
+    pub fn connect(&self, hostname: &str, port: u16) -> Result<Stream> {
```
kyeah: gonna be that guy commenting on docs 😬. maybe document the default options for SSL as well here, to be explicit about what the user should expect?

src/lib.rs
```diff
@@ -179,6 +185,7 @@ impl ClientOptions {
             heartbeat_frequency_ms: DEFAULT_HEARTBEAT_FREQUENCY_MS,
             server_selection_timeout_ms: DEFAULT_SERVER_SELECTION_TIMEOUT_MS,
             local_threshold_ms: DEFAULT_LOCAL_THRESHOLD_MS,
+            stream_connector: StreamConnector::Tcp,
```
kyeah: ::default()?

saghm: I can change it to that

src/lib.rs
```diff
@@ -263,54 +284,54 @@ impl ThreadedClient for Client {
...
+
+                    top.servers.insert(host.clone(), server);
+                }
             }
-        }
 
-        Ok(client)
-    }
+            Ok(client)
+        }
```
saghm: Yeah, I've grown less and less fond of rustfmt after using clang-format quite a bit at work. I'm kind of hoping it matures into something better at some point, but it's not really that close yet.

src/ssl.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+impl Stream for SslStream<TcpStream> {
+    fn peer_addr(&self) -> io::Result<SocketAddr> {
+        self.get_ref().peer_addr()
+    }
+}
+
+pub fn connect(hostname: &str,
```
saghm: Oops, meant to delete this file. Good catch!

src/ssl.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::net::{SocketAddr, TcpStream};
+
+use stream::Stream;
+
+use openssl::ssl::{Ssl, SslMethod, SslContext, SslStream, SSL_OP_NO_COMPRESSION, SSL_OP_NO_SSLV2,
+                   SSL_OP_NO_SSLV3, SSL_VERIFY_NONE, SSL_VERIFY_PEER};
+use openssl::x509::X509_FILETYPE_PEM;
+
+#[derive(Clone, Debug)]
+pub struct SslConfig {
```
saghm: Actually, this file isn't supposed to even exist anymore. As mentioned below, I totally meant to delete it.

src/stream.rs
saghm: No need to apologize, you're absolutely right. I completely forgot about documentation for this PR.

src/stream.rs
saghm: I'm not quite sure I understand the rationale for that. Are you suggesting that we avoid implementing `flush()` for `Stream`? If so, that isn't really feasible, as `flush()` is a required method for implementing `Write`, which is non-optional if we want to use `Stream` as-is in the rest of the code.

saghm:
Btw, given how many iterations this PR has gone through, I think it's a prime candidate for me to finally try out the `merge and squash` option on Github, which I've never actually gotten around to doing. Any objections for me merging it like that (and hopefully not accidentally doing something stupid)?

src/stream.rs
kyeah: This would just shrink the code by enum matching once in `fn mut_stream()`. We'd still have

```rust
fn flush() {
  self.mut_stream().flush()
}
```

and the corollaries for the other trait methods; it would just reduce the number of matches we need to do.

kyeah:
hah! I've been using squash and merge for a while now and it's fantastic -- keeps the history readable at a high level, but links to the PR for more depth. 

Pretty straightforward, just make sure you edit the sub-description before confirming the merge :)

saghm:
Oh, I did not realize it linked to the PR, that's pretty awesome! And yep, I was definitely planning on explicitly setting a good commit message.

src/stream.rs
saghm: I don't really feel that strongly either way about whether this is better or not, but if you'd prefer it, I can take a stab at implementing that (which I think should be possible, although it might involve lifetimes).

saghm: Actually, I don't think this is actually possible; there isn't any feasible return type for the method you're proposing.

kyeah: oh, no problem! My hot take was that there may have been some nice generics magic to do it, but it's a really tiny gain anyways.

README.md
```diff
@@ -18,14 +18,22 @@ Installation
...
 ```
 
+Alternately, you can use the MongoDB driver with SSL support. To do this, you must have OpenSSL installed on your system. Then, enable the `ssl` feature for MongoDB in your Cargo.toml:
+
+```
+[dependencies]
+...
+mongodb = { version = "0.1.8", features = ["ssl"] }
+```
+
```
kyeah: Maybe a quick example here or in the library comment of `src/lib.rs` on how to initialize the client with SSL.

src/stream.rs
saghm: I did a quite check with trying to have it return `T: Read + Write`, but it didn't compile, which fits with my understanding of generics; although different calls to the same method can return different types, a single call to a method should have a single, known, concrete type at compile-time. A traditional OO language can just have the return-type be a superclass here, but of course Rust isn't a traditional OO language. (Not that I'm complaining; I like Rust better this way!)

README.md
saghm: Will do.

saghm:
A changelog is definitely a good idea. As for what specifically the new version should be after this change, are you opposed to bumping to `0.2.0`? Obviously there isn't much of a behavioral change for users not using SSL, but this seems to be a pretty significant step towards maturity for the driver, so I feel like (pre-1.0) minor version jump might be a good fit for it.

I'm not super worried about tagging for the driver, as Cargo/crates.io will be used almost exclusively instead cloning and manually compiling the driver. I suppose it could be useful for users who want to fork the driver to add custom functionality for it though, so I wouldn't be opposed to making sure we tag every new release.

saghm:
FYI the Travis build on my fork for this PR failed due to the test timing a command [taking too long with SSL](https://travis-ci.org/saghm/mongo-rust-driver-prototype#L884). It passed for when the build for this repo ran, so I'm not super worried. I'm guessing this is just the fact that SSL is slightly slower than regular TCP, although it might even just be a coincidence that it happened with SSL and was just a random fluke.

I've restarted the build to see if it happens again, but I'm hoping/expecting that it won't.

kyeah:
yeah I'm cool with `0.2.0` — it kinda breaks away from semantic versioning, but I agree it has some symbolic weight that'll be good for it.

saghm:
Technically, nothing really "breaks away" from semantic versioning when you're pre-1.0 :)

PR #181, opened 2017-01-23
Add test documentation to README

### LINKED ISSUE #163: Add documentation for how to run tests
#162 exposed a hole in our documentation when it comes to how to properly run our test suite. We should add something along the lines of a TESTING.md (and link to it in the README) detailing how to  properly run them to avoid more confusion.

kyeah:
This (finally) fixes #163. We'll likely need to update this a little bit once SSL support goes out in #175.


README.md
```diff
@@ -74,3 +74,47 @@ fn main() {
...
 ```
+
+Testing
+-------
+
+The driver test suite is largely composed of a mixture of integration tests and behavioral unit-tests, relying on the official [MongoDB specifications repo](https://github.com/mongodb/specifications). The easiest way to thoroughly test the driver is to set your fork up with TravisCI.
+
+To test the driver locally, you'll need to setup integration tests and specification tests.
+
+> NOTE: Each integration test uses a unique database/collection to allow tests to be parallelized, and will drop their dependencies before running. However, effects are _not_ cleaned up afterwards.
```
saghm: Good idea to mention this; I don't think I would have thought of it

README.md
```diff
@@ -74,3 +74,47 @@ fn main() {
...
+
+#### Setting up integration tests
+
+All integration tests run on the default MongoDB port, 27017. Before running the tests, ensure that you setup a test database on that port.
+
+Download a version of MongoDB:
```
saghm: Maybe add a "if you don't already have this installed" parenthetical? I'm guessing the majority of people wanting to this this will be doing it just to verify that that it works on their specific system.

README.md
```diff
@@ -74,3 +74,47 @@ fn main() {
...
+```
+wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.2.11.tgz
```
saghm: It's probably better to just point users to [https://www.mongodb.com/download-center](https://www.mongodb.com/download-center) so they can download whichever release they need rather than specifying the Linux link directly.

README.md
```diff
@@ -74,3 +74,47 @@ fn main() {
...
+tar xvf mongodb-linux-x86_64-3.2.11.tgz
+mv mongodb-linux-x86_64-3.2.11 3.2.11
+```
+
+Run a MongoDB server on 27017:
+```
+mkdir -p ./data/test_db_3_2
+3.2.11/bin/mongod --fork --nopreallocj --dbpath ./data/test_db_3_2 --syslog --port 27017
```
saghm: I'd change this to just `mongod --dbpath ./data/test_db_3_2`; I'm guessing most users running the tests won't want to continue running the server in the background after the tests are finished, `--syslog` is only necessary if you're using `--fork`, and `--nopreallocj` doesn't do anything on WiredTiger, which is the default in 3.2. (On 3.0 and earlier, it's only really worth if if the startup time of the database is too long for you for some reason, but given that the tests will likely end up needing to allocate the data files anyways, it's probably better not to use it there either. I took out the option in the Travis config file in #175 for similar reasons, as explained [here](https://github.com/mongodb-labs/mongo-rust-driver-prototype/pull/175/files#r94277292)).

Also, similar to above, it might be worth mentioning that you can just run `mongod ...` instead of `3.2.11/bin/mongod` if you already have it installed.

README.md
```diff
@@ -74,3 +74,47 @@ fn main() {
...
+```
+
+See a full list of versions being tested on Travis in [the travis config](/.travis.yml).
```
saghm: Is `/.travis.yml` a type of `./travis.yml`, or is `/` intended to mean the repository root? If the former case, it should be fixed, and if the latter, I'd suggest something a bit clearer such as `$reporoot/.travis.yml` or `path/to/repository/.travis.yml`.

kyeah: `/` is the generalized root; the link will go to the file page if on github, and to the local version for people looking at the README locally. This is for those who would want to follow the link to the local version, e.g. if they're on a branch with an updated travis config.

saghm: Ah, okay, I completely missed that this was inside a link. Ignore my irrelevant comment.

Issue #179, opened 2017-01-23
3.4 compatibility


saghm:
A few new features were added to MongoDB in 3.4 which we haven't implemented support for yet. The release notes of 3.4 are [here](https://docs.mongodb.com/manual/release-notes/3.4/), but not everything there requires a change in drivers. Off the top of my head, I know that views, the decimal type, and likely collations will require additions/changes to the code.

Additionally, three tests currently fail when run against 3.4. Two of them (`apm::logging` and `client::db::create_collection` seem to be due to the same issue, namely an option that no longer seems to be supported, and the other (`client::db::create_and_get_users`) seems like it might be an issue with the ordering of the results. I'm guessing all of these are issues with the tests rather than the actual driver; if this is the case, we'll likely have to do the same kind of fix we did with 3.0 vs.3.2.

saghm:
Progress update: #185 fixed a number of the issues the driver had with 3.4.x and fixed all of the tests failing except for `client::db::create_and_get_users`. We'll still have to add views, collations, and the decimal type, the last of which will probably only require changes to `bson` (https://github.com/zonyitoo/bson-rs/issues/53).

saghm:
Another update (because I forgot to reference this issue in the PR): #188 fixed the tests for 3.4, so that leaves only implementing the new features.

Iulian7:
Is there any progress on these features?

saghm:
Unfortunately there hasn't been yet. I hope to find some time soon to dig through the backlog of features we have planned for this driver.

saghm:
Hello everyone! We've just released an alpha of [an official MongoDB Rust driver](https://github.com/mongodb/mongo-rust-driver); you can read about it [here](https://www.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver)! Because of this, we won't be updating this driver anymore, and the repository will be made read-only as soon as I get finished posting these messages. Thanks for using this driver, and we hope you try out the new one!

PR #185, opened 2017-02-03
remove specifications of default options

saghm:
Replaces #183 and moves us a bit closer towards #179 

This is a big one; some of the changes of the options structs were relatively easy to integrate, but others (I'm looking you, `FindOptions`) required quite a bit of finagling with parts of the code I hadn't looked closely at in a while.

All tests pass on 3.2, and the only two failing tests on 3.4 are `apm::logging` (I'll need to look into this more closely) and `client::db::create_and_get_users` (which seems to return the users in a different order than before), both of which appear to be issues with the tests not being compatible with 3.4 rather than the driver itself. This means the driver should work with 3.4 after this is merged!

### DIFF
--- a/src/coll/batch.rs
+++ b/src/coll/batch.rs
@@ -1,7 +1,7 @@
 //! Models for collection-level batch operations.
 use super::options::WriteModel;
 
-use bson::Document;
+use bson::{Bson, Document};
 use std::convert::From;
 
 #[derive(Debug)]
@@ -10,33 +10,54 @@ pub struct DeleteModel {
     pub multi: bool,
 }
 
-#[derive(Debug)]
-pub struct UpdateModel {
-    pub filter: Document,
-    pub update: Document,
-    pub upsert: bool,
-    pub multi: bool,
-    pub is_replace: bool,
-}
-
-impl UpdateModel {
-    pub fn new(filter: Document, update: Document, upsert: bool, multi: bool) -> UpdateModel {
-        UpdateModel {
-            filter: filter,
-            update: update,
-            upsert: upsert,
-            multi: multi,
-            is_replace: false,
-        }
-    }
-}
-
 impl DeleteModel {
     pub fn new(filter: Document, multi: bool) -> DeleteModel {
         DeleteModel {
             filter: filter,
             multi: multi,
         }
+    }
+}
+
+#[derive(Debug)]
+pub struct UpdateModel {
+    pub filter: Document,
+    pub update: Document,
+    pub upsert: Option<bool>,
+    pub multi: bool,
+}
+
+impl UpdateModel {
+    pub fn new(filter: Document,
+               update: Document,
+               upsert: Option<bool>,
+               multi: bool)
+               -> UpdateModel {
+        UpdateModel {
+            filter: filter,
+            update: update,
+            upsert: upsert,
+            multi: multi,
+        }
+    }
+}
+
+impl From<UpdateModel> for Document {
+    fn from(model: UpdateModel) -> Self {
+        let mut document = doc! {
+            "q" => (model.filter),
+            "u" => (model.update)
+        };
+
+        if let Some(upsert) = model.upsert {
+            document.insert("upsert", Bson::Boolean(upsert));
+        }
+
+        if model.multi {
+            document.insert("multi", Bson::Boolean(true));
+        }
+
+        document
     }
 }
 
@@ -69,7 +90,6 @@ impl From<WriteModel> for Batch {
                                        update: update,
                                        upsert: upsert,
                                        multi: false,
-                                       is_replace: true,
                                    }])
             }
             WriteModel::UpdateOne { filter, update, upsert } => {
@@ -78,7 +98,6 @@ impl From<WriteModel> for Batch {
                                        update: update,
                                        upsert: upsert,
                                        multi: false,
-                                       is_replace: false,
                                    }])
             }
             WriteModel::UpdateMany { filter, update, upsert } => {
@@ -87,7 +106,6 @@ impl From<WriteModel> for Batch {
                                        update: update,
                                        upsert: upsert,
                                        multi: true,
-                                       is_replace: false,
                                    }])
             }
         }
@@ -148,7 +166,6 @@ impl Batch {
                             update: update,
                             upsert: upsert,
                             multi: false,
-                            is_replace: true,
                         })
                     }
                     WriteModel::UpdateOne { filter, update, upsert } => {
@@ -157,7 +174,6 @@ impl Batch {
                             update: update,
                             upsert: upsert,
                             multi: false,
-                            is_replace: false,
                         })
                     }
                     WriteModel::UpdateMany { filter, update, upsert } => {
@@ -166,7 +182,6 @@ impl Batch {
                             update: update,
                             upsert: upsert,
                             multi: true,
-                            is_replace: false,
                         })
                     }
                     _ => return Some(model),
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -13,7 +13,7 @@ use self::options::*;
 use self::results::*;
 
 use ThreadedClient;
-use common::{ReadPreference, WriteConcern};
+use common::{merge_options, ReadPreference, WriteConcern};
 use cursor::Cursor;
 use db::{Database, ThreadedDatabase};
 
@@ -96,24 +96,32 @@ impl Collection {
                      pipeline: Vec<bson::Document>,
                      options: Option<AggregateOptions>)
                      -> Result<Cursor> {
-        let opts = options.unwrap_or_else(AggregateOptions::new);
-
-        let pipeline_map = pipeline.iter()
-            .map(|bdoc| Bson::Document(bdoc.to_owned()))
+        let pipeline_map: Vec<_> = pipeline.into_iter()
+            .map(|bdoc| Bson::Document(bdoc))
             .collect();
 
-        let mut spec = bson::Document::new();
-        let mut cursor = bson::Document::new();
-        cursor.insert("batchSize", Bson::I32(opts.batch_size));
-        spec.insert("aggregate", Bson::String(self.name()));
-        spec.insert("pipeline", Bson::Array(pipeline_map));
-        spec.insert("cursor", Bson::Document(cursor));
-        if opts.allow_disk_use {
-            spec.insert("allowDiskUse", Bson::Boolean(opts.allow_disk_use));
-        }
+        let mut spec = doc! {
+            "aggregate" => (self.name()),
+            "pipeline" => pipeline_map
+        };
 
-        let read_pref = opts.read_preference.unwrap_or(self.read_preference.to_owned());
-        self.db.command_cursor(spec, CommandType::Aggregate, read_pref)
+        let mut read_preference = self.read_preference.clone();
+
+        match options {
+            Some(aggregate_options) => {
+                if let Some(ref read_preference_option) = aggregate_options.read_preference {
+                    read_preference = read_preference_option.clone();
+                }
+
+                spec = merge_options(spec, aggregate_options);
+            }
+            None => {
+                let cursor = doc! { "batchSize" => (::cursor::DEFAULT_BATCH_SIZE) };
+                spec.insert("cursor", cursor);
+            }
+        };
+
+        self.db.command_cursor(spec, CommandType::Aggregate, read_preference)
     }
 
     /// Gets the number of documents matching the filter.
@@ -121,28 +129,27 @@ impl Collection {
                  filter: Option<bson::Document>,
                  options: Option<CountOptions>)
                  -> Result<i64> {
-        let opts = options.unwrap_or_else(CountOptions::new);
-
         let mut spec = bson::Document::new();
         spec.insert("count", Bson::String(self.name()));
-        spec.insert("skip", Bson::I64(opts.skip as i64));
-        spec.insert("limit", Bson::I64(opts.limit));
-        if filter.is_some() {
-            spec.insert("query", Bson::Document(filter.unwrap()));
+
+        if let Some(filter_doc) = filter {
+            spec.insert("query", Bson::Document(filter_doc));
         }
 
-        // Favor specified hint document over string
-        if opts.hint_doc.is_some() {
-            spec.insert("hint", Bson::Document(opts.hint_doc.unwrap()));
-        } else if opts.hint.is_some() {
-            spec.insert("hint", Bson::String(opts.hint.unwrap()));
+        let mut read_preference = self.read_preference.clone();
+
+        if let Some(count_options) = options {
+            if let Some(ref read_preference_option) = count_options.read_preference {
+                read_preference = read_preference_option.clone();
+            }
+
+            spec = merge_options(spec, count_options);
         }
 
-        let read_pref = opts.read_preference.unwrap_or(self.read_preference.to_owned());
-        let result = try!(self.db.command(spec, CommandType::Count, Some(read_pref)));
+        let result = try!(self.db.command(spec, CommandType::Count, Some(read_preference)));
         match result.get("n") {
-            Some(&Bson::I32(ref n)) => Ok(*n as i64),
-            Some(&Bson::I64(ref n)) => Ok(*n),
+            Some(&Bson::I32(n)) => Ok(n as i64),
+            Some(&Bson::I64(n)) => Ok(n),
             _ => Err(ResponseError(String::from("No count received from server."))),
         }
     }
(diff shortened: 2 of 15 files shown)


src/coll/mod.rs
```diff
@@ -96,53 +96,60 @@ impl Collection {
...
+            spec = merge_options(spec, count_options);
         }
 
-        let read_pref = opts.read_preference.unwrap_or(self.read_preference.to_owned());
-        let result = try!(self.db.command(spec, CommandType::Count, Some(read_pref)));
+        let result = try!(self.db.command(spec, CommandType::Count, Some(read_preference)));
         match result.get("n") {
-            Some(&Bson::I32(ref n)) => Ok(*n as i64),
-            Some(&Bson::I64(ref n)) => Ok(*n),
+            Some(&Bson::I32(n)) => Ok(n as i64),
```
saghm: Numeric types (such as `i32` and `i64`) implement `Copy`, so there isn't any need to capture them by reference.

src/coll/mod.rs
```diff
@@ -153,18 +160,23 @@ impl Collection {
...
-
-        let opts = options.unwrap_or_else(DistinctOptions::new);
-
         let mut spec = bson::Document::new();
         spec.insert("distinct", Bson::String(self.name()));
         spec.insert("key", Bson::String(String::from(field_name)));
-        if filter.is_some() {
-            spec.insert("query", Bson::Document(filter.unwrap()));
+
+        if let Some(filter_doc) = filter {
```
saghm: Although using `is_some()` and `unwrap()` is technically safe, it's more idiomatic to use `if let Some(...)`, so I changed to this style here and a few other places.

src/coll/mod.rs
```diff
@@ -240,34 +252,23 @@ impl Collection {
 
     // Helper method for all findAndModify commands.
     fn find_and_modify(&self,
-                       cmd: &mut bson::Document,
                        filter: bson::Document,
+                       options: bson::Document,
                        _max_time_ms: Option<i64>,
```
saghm: I've noticed we don't seem to actually use this option anywhere in the driver. I think it might be a good idea to look at how other drivers use it, since it probably is included in the spec for *something*.

src/coll/mod.rs
```diff
@@ -276,52 +277,26 @@ impl Collection {
         Ok(doc)
     }
 
-    // Helper method for validated replace and update commands.
-    fn find_one_and_replace_or_update(&self,
```
saghm: After some refactoring to pass the options document directly around, this method isn't necessary anymore

src/coll/mod.rs
```diff
@@ -391,7 +380,6 @@ impl Collection {
                         update: replacement,
                         upsert: upsert,
                         multi: false,
-                        is_replace: true,
```
saghm: This field isn't used anywhere in the driver

src/coll/mod.rs
```diff
@@ -418,7 +404,7 @@ impl Collection {
         vec![Batch::Insert(inserts), Batch::Delete(deletes), Batch::Update(updates)]
     }
 
-    pub fn get_ordered_batches(mut requests: VecDeque<WriteModel>) -> Vec<Batch> {
+    fn get_ordered_batches(mut requests: VecDeque<WriteModel>) -> Vec<Batch> {
```
saghm: I'm unsure why this method was public; it's only used internally

src/coll/mod.rs
```diff
@@ -448,7 +434,7 @@ impl Collection {
             .map(|doc| WriteModel::InsertOne { document: doc.clone() })
             .collect();
 
-        let options = Some(InsertManyOptions::new(ordered, None));
+        let options = Some(InsertManyOptions { ordered: Some(ordered), ..Default::default() });
```
saghm: Isn't this a great syntax? This is a great syntax

src/coll/mod.rs
```diff
@@ -568,7 +554,7 @@ impl Collection {
     // Internal insertion helper function. Returns a vec of collected ids and a possible exception.
     fn insert(&self,
               docs: Vec<bson::Document>,
-              ordered: bool,
+              options: Option<InsertManyOptions>,
```
saghm: Here (and several other places), it made more sense to pass the options struct around directly rather than pulling out the individual fields. In this method it makes little difference, but in some others it simplifies things significantly, so I changed it here for consistency.

src/coll/options.rs
```diff
@@ -1,25 +1,39 @@
 //! Options for collection-level operations.
 use bson::{self, Bson};
-use cursor;
 use common::{ReadPreference, WriteConcern};
 use Error::ArgumentError;
 use Result;
 
 /// Describes the type of cursor to return on collection queries.
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
saghm: A lot of these types didn't have `Debug` implemented for them, which made debugging some things a pain (go figure).

src/coll/options.rs
```diff
@@ -1,25 +1,39 @@
...
+
 /// Describes the type of document to return on write operations.
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ReturnDocument {
     Before,
     After,
 }
 
+impl ReturnDocument {
```
saghm: This was moved from below, where it was called `to_bool`. It's more idiomatic to name methods that take `self` as a reference `as_*` and reserve `to_*` for methods that consume self.

src/coll/options.rs
```diff
@@ -29,58 +43,126 @@ pub enum WriteModel {
...
+
+        if let Some(hint) = options.hint {
+            document.insert("hint", Bson::String(hint));
+        }
+
+        if let Some(hint_doc) = options.hint_doc {
+            document.insert("hint_doc", Bson::Document(hint_doc));
+        }
+
+        // maxTimeMS is not currently used by the driver.
```
saghm: Like I mentioned elsewhere, it seems like there are a few options from the spec that we aren't using. For now, I just documented them in the `Document::from` methods, but we should look into these to see if there's something we should be doing with them.

src/coll/options.rs
```diff
@@ -29,58 +43,126 @@ pub enum WriteModel {
...
 /// Options for collection queries.
-#[derive(Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct FindOptions {
     pub allow_partial_results: bool,
     pub no_cursor_timeout: bool,
-    pub op_log_replay: bool,
-    pub skip: u32,
-    pub limit: i32,
+    pub oplog_replay: bool,
```
saghm: In MongoDB parlance, "oplog" is generally used as a single word, despite the fact that it isn't really a word at all in actual parlance

src/coll/options.rs
```diff
@@ -89,28 +171,134 @@ pub struct FindOptions {
...
     pub projection: Option<bson::Document>,
     pub sort: Option<bson::Document>,
-    pub upsert: bool,
+    pub upsert: Option<bool>,
     pub write_concern: Option<WriteConcern>,
 }
 
+impl FindOneAndUpdateOptions {
+    pub fn new() -> Self {
+        Default::default()
```
saghm: From what I've read, it's considered more idiomatic to implement `new` by calling `default` rather than vice-versa, presumably because `Default` is often derived rather than directly implemented.

src/coll/options.rs
```diff
@@ -134,150 +322,19 @@ pub struct IndexOptions {
...
 /// A single index model.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct IndexModel {
     pub keys: bson::Document,
     pub options: IndexOptions,
 }
 
-/// Options for insertMany operations.
-#[derive(Clone, Default)]
```
saghm: All of these implementations were moved to reside closer to their type definitions, for reasons mentioned above.

src/common.rs
```diff
@@ -95,3 +95,8 @@ impl WriteConcern {
         bson
     }
 }
+
+pub fn merge_options<T: Into<bson::Document>>(document: bson::Document, options: T) -> bson::Document {
```
saghm: Slight tangent: this method exemplifies so much of what I love about Rust. Bounded generics, composable standard library traits, a great iterator API, type inference; it's all here.

src/cursor.rs
```diff
@@ -238,32 +236,24 @@ impl Cursor {
         // Send read_preference to the server based on the result from server selection.
         let new_query = if !send_read_pref {
             query
+        } else if query.get("$query").is_some() {
```
saghm: Given that the pattern match wasn't actually capturing any variables, it seemed more eadable to change it to a pair of `else (if)` clauses.

src/cursor.rs
```diff
@@ -293,45 +280,32 @@ impl Cursor {
...
         let init_time = time::precise_time_ns();
         let result = Message::new_query(req_id,
                                         flags,
-                                        namespace.to_owned(),
-                                        number_to_skip,
-                                        batch_size,
-                                        query.clone(),
-                                        return_field_selector);
+                                        namespace.clone(),
+                                        options.skip.unwrap_or(0) as i32,
```
saghm: I still have no idea why the spec says that the `skip` options should be an i64 if the wire protocol only accepts an i32. Oh, well...

src/db/mod.rs
```diff
@@ -443,19 +422,18 @@ impl ThreadedDatabase for Database {
                  users: Vec<&str>,
                  options: Option<UserInfoOptions>)
                  -> Result<Vec<bson::Document>> {
-        let info_options = options.unwrap_or_else(UserInfoOptions::new);
-        let vec = users.into_iter()
+        let vec: Vec<_> = users.into_iter()
             .map(|user| {
-                let doc = doc! { "user" => user, "db" => (Bson::String(self.name.to_owned())) };
+                let doc = doc! { "user" => user, "db" => (self.name.to_owned()) };
```
saghm: No need to specify the enum variant here, as the `From` trait handles that for us.

src/db/options.rs
```diff
@@ -1,28 +1,59 @@
...
+            document.insert("size", Bson::I64(size));
+        }
+
+        if let Some(max) = options.max {
+            document.insert("max", Bson::I64(max));
+        }
+
+        let mut flags = 0;
+
+        if let Some(true) = options.use_power_of_two_sizes {
```
saghm: Perhaps another place where we could use `bitflags`? Not sure if it's worth it for something so small or not

src/topology/monitor.rs
```diff
@@ -250,7 +250,11 @@ impl Monitor {
 
     /// Returns an isMaster server response using an owned monitor socket.
     pub fn is_master(&self) -> Result<(Cursor, i64)> {
-        let options = FindOptions::new().with_limit(1);
+        let mut options = FindOptions::new();
```
saghm: If we integrate `derive_builder`, this will go back to the way it looked before

tests/apm/mod.rs
```diff
@@ -57,7 +57,8 @@ fn logging() {
     let _ = fs::remove_file("test_apm_log.txt");
 
     // Reset State
-    let reset_client = Client::connect("localhost", 27017).expect("Failed to connect to localhost:27017");
+    let reset_client = Client::connect("localhost", 27017)
```
saghm: The most recent large rustfmt/clippy PR only checked `src` (and not `tests`), so rustfmt picked up a few things.

tests/apm/mod.rs
```diff
@@ -91,8 +92,7 @@ fn logging() {
 
     // Create collection started
     read_first_non_monitor_line(&mut file, &mut line);
-    assert_eq!("COMMAND.create_collection 127.0.0.1:27017 STARTED: { create: \"logging\", \
-                capped: false, auto_index_id: true, flags: 1 }\n",
+    assert_eq!("COMMAND.create_collection 127.0.0.1:27017 STARTED: { create: \"logging\" }\n",
```
saghm: Here (and in several places below) the logging output changed due to the default fields no longer being present.

tests/client/crud_spec/framework.rs
```diff
@@ -170,7 +170,7 @@ macro_rules! run_insert_one_test {
 macro_rules! run_replace_one_test {
     ( $db:expr, $coll:expr, $filter:expr, $replacement:expr, $upsert:expr,
         $outcome:expr ) => {{
-            let options = ReplaceOptions::new($upsert, None);
+            let options = ReplaceOptions { upsert: $upsert, write_concern: None };
```
saghm: I ran into a weird thing here when I tried to use the `..Default::default()` syntax; the compiler seemed to be treating the `..` as part of a range expression. Strangely, it only seemed to occur in certain places in macro definitions.

tests/json/crud/options.rs
```diff
@@ -102,32 +98,26 @@ impl FromJson for FindOneAndUpdateOptions {
     fn from_json(object: &Object) -> FindOneAndUpdateOptions {
         let mut options = FindOneAndUpdateOptions::new();
 
-        let f = |x| Some(Bson::from_json(x));
-        options.projection = match object.get("projection").and_then(f) {
-            Some(Bson::Document(doc)) => Some(doc),
-            _ => None,
-        };
+        if let Some(Bson::Document(projection)) = object.get("projection").map(Bson::from_json) {
```
saghm: I'll be honest; I have absolutely no idea why I implemented this (and the similar cases below) in such a convoluted way originally. I'll just have to chalk it up to my lack of familiarity with Rust at that point.

src/coll/options.rs
```diff
@@ -29,58 +43,126 @@ pub enum WriteModel {
...
+    }
+}
+
 /// Options for count queries.
-#[derive(Clone, Default)]
+#[derive(Clone, Debug, Default)]
 pub struct CountOptions {
-    pub skip: u64,
-    pub limit: i64,
+    pub skip: Option<i64>,
```
saghm: I'm not quite sure why an unsigned integer was used here (and in another place below). The spec doesn't use unsigned integers (probably due to there not being an easy why to enforce it in dynamically typed languages)

src/coll/mod.rs
```diff
@@ -184,35 +196,32 @@ impl Collection {
...
-            doc.insert(String::from("$query"),
-                       Bson::Document(filter.unwrap_or_else(bson::Document::new)));
-
-            doc.insert(String::from("$orderby"),
-                       Bson::Document(options.sort.as_ref().unwrap().clone()));
-
-            doc
-        } else {
-            filter.unwrap_or_else(bson::Document::new)
+        let find_options = options.unwrap_or_else(FindOptions::new);
```
kyeah: verrrry tiny win here, but we can do `options.unwrap_or_default()` here.

src/coll/mod.rs
```diff
@@ -184,35 +196,32 @@ impl Collection {
...
                       flags,
-                      options.skip as i32,
-                      options.limit,
                       doc,
-                      options.projection.clone(),
+                      find_options,
                       cmd_type,
                       false,
-                      read_pref)
+                      read_preference)
```
kyeah: Interesting. I wonder if we can remove this read_pref option here and skip the mandatory cloning? Perhaps something to look into at a later point.

src/coll/mod.rs
kyeah: 🤔

src/coll/mod.rs
kyeah: **nice**

src/coll/mod.rs
kyeah: _nice_

src/coll/mod.rs
```diff
@@ -757,23 +748,20 @@ impl Collection {
 
         let mut updates = Vec::new();
         for model in models {
-            let mut update = bson::Document::new();
-            update.insert("q", Bson::Document(model.filter));
-            update.insert("u", Bson::Document(model.update));
-            update.insert("upsert", Bson::Boolean(model.upsert));
+            let mut update = bson::Document::from(model);
```
kyeah: yes

src/coll/mod.rs
```diff
@@ -757,23 +748,20 @@ impl Collection {
...
 
-        let mut cmd = bson::Document::new();
-        cmd.insert("update", Bson::String(self.name()));
-        cmd.insert("updates", Bson::Array(updates));
-        cmd.insert("writeConcern", Bson::Document(wc.to_bson()));
+        let cmd = doc! {
+            "update" => (self.name()),
+            "updates" => updates,
+            "writeConcern" => (wc.to_bson())
+        };
```
kyeah: _yes_

src/common.rs
kyeah: Would this make more sense as a function on bson::Document? Feels reaal funny to have this global function, and languages like ruby have this type of method on the hash object itself. `doc.merge(other_doc)`

src/db/options.rs
kyeah: oh nice! yeah, it only has two flags, so probably not worth it.

tests/client/crud_spec/framework.rs
kyeah: macros: still sketch.

src/coll/batch.rs
```diff
@@ -10,33 +10,54 @@ pub struct DeleteModel {
     pub multi: bool,
 }
 
+impl DeleteModel {
+    pub fn new(filter: Document, multi: bool) -> DeleteModel {
```
saghm: Seems like it's worth adding that in the bson crate.

src/common.rs
saghm: I was thinking that was well when I was implementing it. I think it's worth making a PR to the bson crate for this, but in the interest of getting the driver working with 3.4, I'd suggest we don't block this merge on it and open another PR once the functionality is added to bson.

src/cursor.rs
```diff
@@ -27,7 +27,8 @@ use {Client, CommandType, Error, ErrorCode, Result, ThreadedClient};
 use apm::{CommandStarted, CommandResult, EventRunner};
 
 use bson::{self, Bson};
-use common::{ReadMode, ReadPreference};
+use common::{ReadMode, merge_options, ReadPreference};
```
saghm: Oops, I actually meant to put it alphabetically, which would be at the beginning. Is that okay too?

tests/client/crud_spec/framework.rs
saghm: Yeah, I think I probably only implemented all this in macros at the time because I had finally gotten to the point where I understood them and therefore wanted to celebrate by using them everywhere.

Let's open an issue to convert the test frameworks from macros to legitimate Rust code; it'll be a bit of work, but it'll probably save a lot of hassle in the long run.

saghm:
Sounds good!

PR #187, opened 2017-02-07
update description

saghm:
@kyeah thoughts? I figure it's easiest to just have any discussion about the wording/content right here in the PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Valeri Karpov <valkar207@gmail.com>",
            "Saghm Rossi <saghmrossi@gmail.com>",
            "Kevin Yeh <kevinyeah@utexas.edu>"]
 
-description = "An experimental MongoDB driver written by MongoDB interns."
+description = "A native MongoDB driver developed under MongoDB labs."
 repository = "https://github.com/mongodb-labs/mongo-rust-driver-prototype"
 documentation = "https://docs.rs/mongodb"
 readme = "README.md"


Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Valeri Karpov <valkar207@gmail.com>",
            "Saghm Rossi <saghmrossi@gmail.com>",
            "Kevin Yeh <kevinyeah@utexas.edu>"]
 
-description = "An experimental MongoDB driver written by MongoDB interns."
+description = "A native-Rust MongoDB driver. Originally written by MongoDB interns, the driver is an unofficial but still actively maintained project."
```
kyeah: > native-Rust

Since you only see this in descriptions, I think the standard is just 'native'. See: [postgres](https://crates.io/crates/postgres)

> Originally written by MongoDB interns.

I don't think this is really relevant information anymore, and it might lead people away. I understand the need to note that it started as an in-house project but is still unofficially supported. Perhaps we can play around with this?

"A native MongoDB driver. Actively maintained, yet unofficially supported for MongoDB 3.0+."

saghm: Hmmm, I was thinking it might be worth mentioning its origins because its the only Rust driver for MongoDB that's at all affiliated with the company. Maybe I should check to see if there's any recommended language that denotes that the driver is owned by MongoDB but not officially supported.

kyeah: perhaps some inspiration from the [Go driver page](https://docs.mongodb.com/ecosystem/drivers/go/) (community-driven) and [mongodb-labs description](https://github.com/mongodb-labs) ("from MongoDB Inc.")?

"A native MongoDB driver. Developed under MongoDB labs and maintained by the community."

saghm: In terms of the copyright, I'm pretty sure it's owned by MongoDB since we wrote it while employed by the company, and in terms of the Github repo, it certainly is! The Go driver is in a bit different situation; the person who wrote it used to be contracted by MongoDB to maintaining it, but that's no longer the case, hence the disparity between "developed under" and "maintained by".

Working on this driver is explicitly not part of my job or related to my employment with MongoDB,  but I'm not exactly maintaining it  "as a community member" in the sense that "community" is generally by MongoDB to describe drivers. Honestly, I think that this driver is in a fairly unique position that the usual terminology that MongoDB uses to describe drivers isn't really that well-equipped to handle, which is part of why I originally suggested just saying exactly what the origins and current state of the driver is, i.e. created by MongoDB interns, not officially supported but actively maintained. If you still aren't fond of keeping the "created by MongoDB interns" part, I'd be all right with something similar to your suggestion in the first comment. I'd want to tweaking the language, as at least to me, it sounds like it's saying the driver doesn't officially support MongoDB 3.0+, which isn't the case; the driver definitely supports MongoDB the database, it's just that MongoDB the company doesn't provide official support for the driver.

kyeah: I think the main points that we're circling around is that:
1. It's associated with/developed under MongoDB, i.e. it's under MongoDB "incubation" via Labs
2. It's not officially supported by MongoDB, which can either mean it's maintained by the "community" (not necessarily) or that it's not production-approved (more closely aligned to what we want to get across).

I don't think there's a reason to differentiate "[originally] developed under" and "maintained by"; I think both of these are underscored by its position as a `mongodb-labs` project, and there's not much of a reason to say that it's "actively maintained" — people can tell the status of a project by it's commit and release history. In the end, I agree that something close to our first two interpretations would be good, but that the semantics can be summarized more succinctly by the connotation of it being a "labs project".

"A native MongoDB driver developed under MongoDB labs."

/shrug

saghm: That works for me; it's concise, correct, and contains all the necessary information.

PR #188, opened 2017-02-22
fix tests for mongodb 3.4

saghm:
Two tests were failing when run against 3.4, both due to the order of things being returned having changed. Along with fixing them, I added 3.4 testing to the travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,23 @@ before_install:
     - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.0.14.tgz
     - tar xvf mongodb-linux-x86_64-ubuntu1204-3.0.14.tgz
     - mv mongodb-linux-x86_64-ubuntu1204-3.0.14 3.0.14
-    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.2.11.tgz
-    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.2.11.tgz
-    - mv mongodb-linux-x86_64-ubuntu1204-3.2.11 3.2.11
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.2.12.tgz
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.2.12.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.2.12 3.2.12
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.4.2.tgz
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.4.2.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.4.2 3.4.2
 
 script:
-    - mkdir -p ./data/db30 ./data/db30-ssl ./data/db32 ./data/db32-ssl
+    - mkdir -p ./data/db30 ./data/db30-ssl ./data/db32 ./data/db32-ssl ./data/db34 ./data/db34-ssl
     - 3.0.14/bin/mongod --fork --dbpath ./data/db30 --syslog --port 27017
     - cargo build --verbose
     - cargo test --verbose
     - killall mongod
-    - 3.2.11/bin/mongod --fork --dbpath ./data/db32 --syslog --port 27017
+    - 3.2.12/bin/mongod --fork --dbpath ./data/db32 --syslog --port 27017
+    - cargo test --verbose
+    - killall mongod
+    - 3.4.2/bin/mongod --fork --dbpath ./data/db34 --syslog --port 27017
     - cargo test --verbose
     - killall mongod
     - 3.0.14/bin/mongod --fork --dbpath ./data/db30 --syslog --port 27017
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,26 +6,34 @@ before_install:
...
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.2.12.tgz
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.2.12.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.2.12 3.2.12
+    - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1204-3.4.2.tgz
+    - tar xvf mongodb-linux-x86_64-ubuntu1204-3.4.2.tgz
+    - mv mongodb-linux-x86_64-ubuntu1204-3.4.2 3.4.2
 
 script:
-    - mkdir -p ./data/db30 ./data/db30-ssl ./data/db32 ./data/db32-ssl
+    - mkdir -p ./data/db30 ./data/db30-ssl ./data/db32 ./data/db32-ssl ./data/db34 ./data/db34-ssl
```
saghm: This is starting to get a bit messy; we should try to look into using mongo orchestration or some other method of managing testing against the different server versions, ideally before 3.6 comes out and we have to add that as well.

tests/apm/mod.rs
```diff
@@ -54,7 +54,7 @@ fn read_first_non_monitor_line(file: &mut BufReader<&File>, line: &mut String) {
 
 #[test]
 fn logging() {
-    let _ = fs::remove_file("test_apm_log.txt");
+    let _ = fs::remove_file("test_log.txt");
```
saghm: For some reason the old name of this file didn't match the name I put in the .gitignore way back when this test was created, so I fixed that.

tests/apm/mod.rs
```diff
@@ -170,5 +180,5 @@ fn logging() {
 
     coll.drop().unwrap();
 
-    fs::remove_file("test_apm_log.txt").unwrap();
+    fs::remove_file("test_log.txt").unwrap();
```
saghm: I've noticed/started to get annoyed that the test file doesn't get cleaned up when the tests fail above this line, so I'm thinking we should change the tests so this file doesn't clutter things up. Given that the file should only be about a dozen lines, I think reading the whole thing into a string and then iterating over the lines should be fine.

tests/client/db.rs
```diff
@@ -35,7 +35,7 @@ fn create_collection() {
 
     let db1 = if v3_1 { "test1" } else { "test2" };
     let db2 = if v3_1 { "test2" } else { "test1" };
-    
+
```
saghm: See, rustfmt does actually help sometimes; empty lines shouldn't have extra whitespace

tests/client/db.rs
```diff
@@ -101,6 +101,17 @@ fn create_and_get_users() {
     db.drop_database().unwrap();
     db.drop_all_users(None).unwrap();
 
+    let kevin_options = CreateUserOptions {
```
saghm: Given that versions below 3.4 return the users in the order they're created and 3.4 returns them in lexicographical order, the easiest fix for this is to just create the users in the order where all versions expect the same thing.

tests/apm/mod.rs
kyeah: That might have been my fault -- i changed a lot of stringy fields to be more specific when I was parallelizing all of the tests.

tests/apm/mod.rs
```diff
@@ -121,10 +124,17 @@ fn logging() {
                 _id: 1 }] }\n",
                &line);
 
+
+    let insert_one_line_start = if v3_3 {
+        "COMMAND.insert_one 127.0.0.1:27017 COMPLETED: { ok: 1, n: 1 } ("
+    } else {
+        "COMMAND.insert_one 127.0.0.1:27017 COMPLETED: { n: 1, ok: 1 } ("
```
kyeah: argh, drives me crazy. this is the expected apm behavior for drivers, yeah? to print it out as ordered by the server, rather than always be lexicographic?

tests/apm/mod.rs
kyeah: Makes sense to me for now. In an ideal future we would allow the client to take in a generic Write-implementing object and pass in a string-like object for testing purposes; if `with_log_file(str)` is part of the spec, it can be added as a utility method.

tests/client/db.rs
kyeah: a rare win, but still on thin ice. _i'm watching you, rustfmt_

tests/client/db.rs
kyeah: nice.

tests/client/db.rs
```diff
@@ -101,6 +101,17 @@ fn create_and_get_users() {
...
+        custom_data: None,
+        roles: vec![Role::All(AllDatabaseRole::Read)],
+        write_concern: None,
+    };
+
+    db.create_user("kevin",
+                     "ihavenosenseofhumorandthereforeihatepuns!",
```
kyeah: lol

.travis.yml
kyeah: Added a tracking issue for Morc (this is a thing now. i'm making this name happen)

tests/apm/mod.rs
saghm: I'll double check, but I'd guess that the spec doesn't even mention it

tests/client/db.rs
saghm: Haha, I had forgotten about that until I made this change, but I'm not even slightly surprised that  did this

tests/apm/mod.rs
kyeah: yeah, makes sense as long as we're supposed to print document fields in insert order (most likely) and read server responses in insert order (probably doesn't even mention this in any of the documentation, but we wouldn't want to do any extra work here to alphabetize anyways).

tests/apm/mod.rs
saghm: That's a good idea; it would be much more idiomatic that way.

saghm:
I'm a bit confused about why Github shows Travis as "not yet completed" when the task clearly finished last night...

PR #192, opened 2017-02-25
Goodbye rustc_serialize, Hello serde_json

### LINKED ISSUE #186: Remove rustc_serialize dependency, Upgrade bson lib
:wave: we should move over to using `serde_json` and `data-encoding` like the bson lib, then upgrade that dependency and shuttle out `rustc_serialize` since it's not really being maintained anymore. See https://github.com/zonyitoo/bson-rs/pull/54.

kyeah:
This fixes #186 and upgrades to the new bson version.

cc. @zonyitoo


Cargo.toml
```diff
@@ -1,36 +1,40 @@
 [package]
-name = "mongodb"
-version = "0.2.2"
-authors = ["Kevin Yeh <kevinyeah@utexas.edu>",
-           "Saghm Rossi <saghmrossi@gmail.com>",
-           "Valeri Karpov <valkar207@gmail.com>"]
-
+authors = ["Kevin Yeh <kevinyeah@utexas.edu>", "Saghm Rossi <saghmrossi@gmail.com>", "Valeri Karpov <valkar207@gmail.com>"]
```
saghm: ...sure, okay

src/auth.rs
```diff
@@ -108,31 +101,22 @@ impl Authenticator {
                                                       String,
                                                       u32);
 
-        let rnonce_b64 = match rnonce_opt {
-            Some(val) => val,
-            None => return Err(ResponseError(String::from("Invalid rnonce returned"))),
-        };
+        let rnonce_b64 = rnonce_opt
```
saghm: Nice

tests/client/crud_spec/framework.rs
```diff
@@ -298,7 +298,7 @@ macro_rules! check_coll {
         };
 
         let coll = match outcome_coll.name {
-            Some(ref str) => $db.collection(&str),
+            Some(ref s) => $db.collection(&s.clone()[..]),
```
saghm: Cloning here feels a little weird. Is there a reason why this needed to be changed from from `&s` to `&s.clone()[..])`?

src/auth.rs
```diff
@@ -108,31 +101,22 @@ impl Authenticator {
...
+            .ok_or(ResponseError(String::from("Invalid rnonce returned")))?;
```
zonyitoo: I would recommend to use `ok_or_else`.

```
.ok_or_else(|_| ResponseError(String::from("Invalid rnonce returned")))
```

src/auth.rs
```diff
@@ -108,31 +101,22 @@ impl Authenticator {
...
-            Some(val) => val,
-            None => return Err(ResponseError(String::from("Invalid salt returned"))),
-        };
-
-        let salt = match salt_b64.from_base64() {
-            Ok(val) => val,
-            Err(_) => return Err(ResponseError(String::from("Invalid base64 salt returned"))),
-        };
+        let salt_b64 = salt_opt
+            .ok_or(ResponseError(String::from("Invalid salt returned")))?;
```
zonyitoo: The same as above

src/auth.rs
```diff
@@ -108,31 +101,22 @@ impl Authenticator {
...
 
+        let salt = base64::decode(salt_b64.as_bytes())
+            .or(Err(ResponseError(String::from("Invalid base64 salt returned"))))?;
```
zonyitoo: Here.

src/auth.rs
```diff
@@ -108,31 +101,22 @@ impl Authenticator {
...
 
-        let i = match i_opt {
-            Some(val) => val,
-            None => return Err(ResponseError(String::from("Invalid iteration count returned"))),
-        };
+        let i = i_opt
+            .ok_or(ResponseError(String::from("Invalid iteration count returned")))?;
```
zonyitoo: And here

src/auth.rs
saghm: Good call, I forgot about that one. I remember clippy suggests that; I wonder if it's possible to set up clippy as a CI tool to check PRs?

Cargo.toml
kyeah: `cargo add`. ¯\\\_(ツ)_/¯

tests/client/crud_spec/framework.rs
kyeah: ah! must have been something weird tripping up the compiler in the intermediary state, while I was transitioning everything to serde_json. will fix.

src/auth.rs
saghm: I wonder if it might be worth figuring out if there are any clippy lints we don't care about, fixing all the others, and then requiring that clippy passes on any future PR (which we could potentially do by just having clippy run as part of the Travis build). Not sure how feasible that would actually be, though...

kyeah: i think it's definitely a good thing to have, and we can incrementally build up a list of clippy rules we want hard style checks on; the only problem I have is keeping up with new or changing clippy rules, but I think the benefits outweigh the upkeep.

saghm:
Do you have any idea what's going on with the Travis build? The "push" one is looking for a branch that doesn't exist, so I don't think we care about that one, but the "pr" one seems to be having issues with the SSL tests (i.e. the server starts in SSL mode fine, but then the tests are unable to connect to it).

kyeah:
There was an accidental push to the mongodb-labs repo, so I think the PR is stuck trying to find that commit (at least for the push check).

 The connection issue is intermittent, and we're having this issue on master as well. I have a branch open to test mongo versions on separate travis jobs, which seems to clear the issue, but I'm not sure what the core issue is.

saghm:
Hmmm...is it worth merging the change to use separate travis jobs into master to fix the issue?

kyeah:
@saghm cool to merge this PR since the `pr` build is passing? I can follow-up with a PR on these intermittent test failures.

saghm:
Works for me

PR #195, opened 2017-02-27
Test mongodb versions in separate Travis jobs

kyeah:
#### what

This PR extracts our test setup into a bash script that runs the full test suite for a single mongodb version; separate Travis jobs run in parallel for each version.

I decided not to cache the mongodb release directory, since it only takes about 2 seconds to download and extract. This allows us to have a global cargo-dependency cache for all of our jobs.

### why parallelize?

Although this doesn't provide much of a performance benefit for successful builds, this will speed up failure notifications — we'll only have to wait for connection timeouts on a single test suite, instead of running through all of them in series.

This also seems to fix our intermittent connection issues when running our test suite. 

The downside to this is that we need to add a new Travis job for each tested mongodb version, and will probably hit the 5-parallel-job limit on our repo at some point.


kyeah:
hm; I tested this on my own fork and had it running 5 concurrent jobs, but it looks like mongodb-labs is only setup for a single job.

.travis.yml
```diff
@@ -2,38 +2,17 @@ language: rust
...
-    - cargo build --features ssl --verbose
-    - cargo test --features ssl --verbose
-    - killall mongod
-    - 3.2.12/bin/mongod --fork --dbpath ./data/db32 --syslog --port 27017
-    - 3.2.12/bin/mongod --fork --dbpath ./data/db32-ssl --syslog --port 27018 --sslMode requireSSL --sslPEMKeyFile tests/ssl/server.pem --sslCAFile tests/ssl/ca.pem
-    - cargo test --features ssl --verbose
-    - killall mongod
-    - 3.4.2/bin/mongod --fork --dbpath ./data/db34 --syslog --port 27017
-    - 3.4.2/bin/mongod --fork --dbpath ./data/db34-ssl --syslog --port 27018 --sslMode requireSSL --sslPEMKeyFile tests/ssl/server.pem --sslCAFile tests/ssl/ca.pem
     - cargo test --features ssl --verbose
```
saghm: I might be misunderstanding, but this looks like we're only testing with SSL enabled? We definitely should be testing both with and without SSL enabled; otherwise, we might accidentally do something like make a change that renders the code with SSL disabled unable to be compiled.

saghm:
I'll check with IT to see if this is something we can get enabled for the repository.

saghm:
Ah, I think I found the issue without having to contact IT; according to the second answer [here](https://stackoverflow.com/questions/29567063/how-does-travis-ci-org-throttle-builds), the "five concurrent jobs" is per entire user/organization, not per repo. When run on one of our personal accounts, we had no other jobs running for our other repos, meaning we could use all our jobs for the Rust driver, whereas there were likely already other jobs running for mongodb-labs (or possibly even just mongodb; I'm not sure if Travis counts them together if they're owned by the same accounts).

Regardless, I think this PR is still worth merging (after the requested changes) both because it seems to have fixed the random failures to connect and because it still *might* be able to run the jobs in parallel if there isn't anything else running. (Maybe we could just make all our PR's at 3 AM just for the enjoyment of them running in parallel).

PR #198, opened 2017-03-11
Integrate clippy lint as a feature

kyeah:
This integrates clippy as a driver feature that can be run on nightly via `cargo build --features lint`. I'm using `lint` to allow it to include other lints in the future (i.e. standard linting).

I went through the [clippy lints](https://github.com/Manishearth/rust-clippy#lints) that are default-allowed, and upped their level to `warn` if there were no warnings, or a few that were easily fixable. A small handful of them seemed okay to allow:

```
cast_possible_truncation
cast_possible_wrap
cast_sign_loss
float_arithmetic
integer_arithmetic
string_add
string_add_assign
```

A couple of these introduced a fair number of warnings or required further discussion, and I've allowed them for now:

```
missing_docs_in_private_items
option_unwrap_used
result_unwrap_used
pub_enum_variant_names
single_match_else
stutter
use_debug
```

There are very few lints that default to `deny`. It might be worth it to review the warnings for egregious lints that we want to block builds on (we may only want to deny them on the CI build via some harsh_lint configuration).

This does not update the CI build to do linting.


kyeah:
As a note, there is one outstanding warning left in:

screen shot 2017-03-11 at 3 10 24 pm

I think the best option here is to remove the `to_bson_array` method; I don't know if we need this utility method. Thoughts?

src/gridfs/file.rs
```diff
@@ -288,18 +288,18 @@ impl File {
     // Retrieves a binary file chunk and asynchronously pre-loads the next chunk.
     fn get_chunk(&mut self) -> Result<Vec<u8>> {
         let id = self.doc.id.clone();
-        let chunk_num = self.chunk_num;
+        let curr_chunk_num = self.chunk_num;
```
saghm: Jus to make sure I understand, the lint here was to remove the shadowed variable name?

saghm:
With regards to `Role::to_bson_array`: the only place it seems to be used is in the implementation of `From<CreateUserOptions> for Document` in `src/db/options.rs`, so I'm fine with just removing the helper method and manually doing the conversion there.

src/gridfs/file.rs
kyeah: yup 👍

kyeah:
> I notice that all the changes are to the driver itself (i.e. not the tests). Did clippy just not lint them, or were there no issues?

ah, you'd have to add these cfgs to `tests/lib.rs`. i'll add them and fix the lints; not sure if there's a good way to centralize these settings :/

tests/client/gridfs.rs
```diff
@@ -34,12 +34,12 @@ fn put_get() {
     let (fs, _, fschunks) = init_gridfs("test-client-gridfs-grid_put");
 
     let name = "grid_put_file";
-    let src_len = (DEFAULT_CHUNK_SIZE as f32 * 2.5) as usize;
+    let src_len = (DEFAULT_CHUNK_SIZE as f64 * 2.5) as usize;
```
saghm: This is an interesting one. I guess it makes sense to use f64 if there's not a reason not to to avoid losing precision, but it doesn't really matter here that much. (If anything,  `DEFAULT_CHUNK_SIZE * 5 / 2` might even be a bit more clean, although I'd want to verify that it doesn't overflow since DEFAULT_CHUNK_SIZE is an i32).

tests/client/wire_protocol.rs
```diff
@@ -22,7 +22,7 @@ fn insert_single_key_doc() {
 
             let cm = match res {
                 Ok(message) => message,
-                Err(_) => panic!("Could not create message!"),
+                _ => panic!("Could not create message!"),
```
saghm: I find it interestint that clippy prefers this; to me, explicitly putting `Err` is much more clear. More generally, I feel like this would help catch potential logic errors if an enum has a new variant added to it (this of course isn't going to happen with Result, but I'm assuming this lint applies to all enums and not just Result.

Is this lint something you feel strongly about? My personal preference would be not to include this one.

tests/json/crud/arguments.rs
```diff
@@ -56,9 +56,7 @@ impl Arguments {
     pub fn aggregate_from_json(object: &Map<String, Value>) -> Result<Arguments, String> {
         let options = AggregateOptions::from_json(object);
 
-        let f = |x| Some(Bson::from_json(x));
-
-        let array = val_or_err!(object.get("pipeline").and_then(f),
+        let array = val_or_err!(object.get("pipeline").map(Bson::from_json),
```
saghm: Oh geez, this is much better. I thought I had cleaned all this up in the PR for removing default options from being sent to the server, but I guess not

tests/json/eq.rs
```diff
@@ -6,6 +6,7 @@ pub trait NumEq {
     fn int_eq(&self, i: i64) -> bool;
 }
 
+#[cfg_attr(feature = "clippy", allow(cast_precision_loss))]
```
saghm: Slightly tangential, but this has been on my radar for a while: we're currently using the `nalgebra` crate to handle floating-point equality, but that's a fairly heavyweight dependency for such a small task. There are several crates for comparing floating points that aren't part of entire linear algebra toolkits; from a quick search, I've found [`approx`](https://crates.io/crates/approx) and [`float-cmp`](https://crates.io/crates/float-cmp). We should look into replacing `nalgebra` with something more fitting at some point.

Cargo.toml
```diff
@@ -23,6 +23,10 @@ semver = "0.5.0"
 separator = "0.3.1"
 time = "0.1.35"
 
+[dependencies.clippy]
+optional = true
+version = "0.0.118"
```
saghm: I didn't think of this until now, but clippy might be one of those rare cases in which we do actually want `version = "*"`; it seems to update fairly often (including once already since this PR was opened). I know this technically mean that we could update to a version with breaking changes, but I've been under the impression that clippy basically does incremental versioning (i.e. that the two leading zeros are superfluous and only present because crates requires a semver-looking version). Plus, the worst case scenario here is that our travis build is temporarily broken; this won't case any errors for users of the library.

kyeah: hm, I'm a little opposed since I'd rather have predictably passing travis builds, instead of either a. waiting for fixes or b. merging PRs with stylistic issues. If new versions bubble up new warnings, it will also impede people who want to make PRs on unrelated things. 

It requires more upkeep to update it constantly, though, so I would be cool with it. Do you know if new versions can bring up warnings on previously-passing code?

saghm: My guess would be that independent of bug fixes (which isn't easy to predict based on the utter contempt for semver in clippy's versioning), only new lints would be added rather than the scope of existing lints being expanded; this is only my intuition though, as I doubt that clippy gives any sort of guarantees about that.

You make a good point about the potential for accidentally mixing lint fixes with other PRs. While I think there might be ways to organize around this, if you'd prefer not to do this then I'm all right with forgoing it. That being said, it might be work figuring out some sort of automated way to of incorporating `cargo outdated` into our workflow (either just for clippy or all of our root dependencies). Of course, to correctly do that, we'd need to figure out how often and when we want to update our dependencies. For some dependencies (e.g. `bitflags`), it's probably not very high priority to update every time a new version comes out, but for things like `bson`, we'd ideally would want to put in the effort to try to keep the version update (like we've been doing lately by releasing a new version whenever `bson` does).

tests/json/crud/arguments.rs
saghm: Ohhh, was this not a clippy thing? I would have been fairly impressed if clippy was able to deal with a closure and a method chain inside  macro in separate statements.

tests/client/wire_protocol.rs
saghm: Once this gets fixed, LGTM

Cargo.toml
kyeah: I'm going to set this to update automatically and see how much trouble it gives us. At any rate, it'll give us a better idea of what types of changes come in clippy updates.

PR #202, opened 2017-04-15
fixes to pass updated spec tests

### LINKED ISSUE #180: Test fails with updated specifications tests
We haven't updated the specifications submodule in a while, so I tried out updating it to see if anything changed. After updating it, one test (`server_selection::replicasetnoprimary::server_selection_replica_set_no_primary_read`) fails with the error "server must have tags" on both 3.2 and 3.4 (I haven't tried 3.0).

We should fix this test and commit the updated submodule.

saghm:
This fixes #180. There were two issues; first, some new JSON spec tests didn't have a field that existing tests of the same type, and second, `Topology_description::choose_hosts` mistakenly returned servers of type 'Unknown' as part of the short-circuit logic if the read mode was 'nearest' (which I understand to be a bug from the description of "Available" in the [server selection spec](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#terms)). Unfortunately, the latter required an API change (namely wrapping the return type in `Result`, as accessing the type of a server requires acquiring a lock (which I assume is because it can get changed in the background monitoring thread).

As usual, this PR comes with a number of "helpful" contributions from rustfmt, this time evidently due to change in the way rustfmt handles indentation.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -80,12 +80,12 @@ impl FromStr for TopologyType {
     type Err = Error;
     fn from_str(s: &str) -> Result<Self> {
         Ok(match s {
-            "Single" => TopologyType::Single,
-            "ReplicaSetNoPrimary" => TopologyType::ReplicaSetNoPrimary,
-            "ReplicaSetWithPrimary" => TopologyType::ReplicaSetWithPrimary,
-            "Sharded" => TopologyType::Sharded,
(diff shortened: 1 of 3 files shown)


saghm:
No worries! I accidentally forgot to assign you to review until four days after submitting it, so you didn't actually take very long after that happened.

PR #207, opened 2017-04-30
Bump bson to 0.7.0

anderspitman:


anderspitman:
When running the example code from the README I got the following error:

```rust
let mut cursor = collection.find(Some(doc.clone()), None)
   |                                           ^^^^^^^^^^^ expected struct `bson::ordered::OrderedDocument`, found a different struct `bson::ordered::OrderedDocument`
   |
   = note: expected type `bson::ordered::OrderedDocument` (struct `bson::ordered::OrderedDocument`)
              found type `bson::ordered::OrderedDocument` (struct `bson::ordered::OrderedDocument`)
note: Perhaps two different versions of crate `bson` are being used?
```

Judging from [this SO answer](http://stackoverflow.com/a/31130554/943814) I solved the problem by bumping the bson version in my `Cargo.toml` to 0.7.0.

saghm:
Good catch! Thanks for the PR

PR #210, opened 2017-05-05
fix cursor documentation

saghm:
I noticed that our documentation in the cursor file was wildly inaccurate, presumably due to not updating it when we made changes.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -40,25 +40,20 @@ pub const DEFAULT_BATCH_SIZE: i32 = 20;
 
 /// Maintains a connection to the server and lazily returns documents from a
 /// query.
-///
-/// # Fields
-///
-/// `client` - The client to read from.
-/// `namespace` - The namespace to read and write from.
-/// `batch_size` - How many documents to fetch at a given time from the server.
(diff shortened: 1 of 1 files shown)


src/cursor.rs
```diff
@@ -40,25 +40,20 @@ pub const DEFAULT_BATCH_SIZE: i32 = 20;
...
 pub struct Cursor {
+    // The client to read from.
     client: Client,
+    // The namespace to read and write from.
     namespace: String,
+    // How many documents to fetch at a given time from the server.
     batch_size: i32,
+    // Uniquely identifies the cursor being returned by the reply.
     cursor_id: i64,
+    // An upper bound on the total number of documents this cursor
```
kyeah: missing the second line here.

src/cursor.rs
```diff
@@ -40,25 +40,20 @@ pub const DEFAULT_BATCH_SIZE: i32 = 20;
...
     limit: i32,
+    // How many documents have been returned so far.
     count: i32,
+    // A cache for documents received from the query that have not
```
kyeah: also here!

PR #211, opened 2017-05-06
fix default batch size

### LINKED ISSUE #209: Default batch size is incorrect
Using this driver for a basic query, the results are being returned to my client much slower than doing the same query with `pymongo`. My best guess is that the rust driver is making a bunch of round trips to the database to retrieve the results a bit at a time, while `pymongo` gets everything in one shot. Again, this is just a guess. If I'm right is there a way to change the behavior. Or is it something else altogether?

saghm:
Fixes #209.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -36,7 +36,7 @@ use wire_protocol::operations::Message;
 
 use std::collections::vec_deque::VecDeque;
 
-pub const DEFAULT_BATCH_SIZE: i32 = 20;
+pub const DEFAULT_BATCH_SIZE: i32 = 0;
 
 /// Maintains a connection to the server and lazily returns documents from a
 /// query.
@@ -462,9 +462,12 @@ impl Cursor {
     /// # Return value
     ///
     /// Returns a vector containing the BSON documents that were read.
-    pub fn next_batch(&mut self) -> Result<Vec<bson::Document>> {
-        let n = self.batch_size;
-        self.next_n(n)
+    pub fn get_current_batch(&mut self) -> Result<Vec<bson::Document>> {
+        if self.buffer.is_empty() {
+            self.get_from_stream()?;
+        }
+
+        Ok(self.buffer.drain(..).collect())
     }
 
     /// Checks whether there are any more documents for the cursor to return.
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -36,6 +36,7 @@ use wire_protocol::operations::Message;
 
 use std::collections::vec_deque::VecDeque;
 
+// Allows the server to decide the batch size.
 pub const DEFAULT_BATCH_SIZE: i32 = 0;
 
 /// Maintains a connection to the server and lazily returns documents from a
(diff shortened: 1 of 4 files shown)


src/cursor.rs
```diff
@@ -469,9 +469,12 @@ impl Cursor {
     /// # Return value
     ///
     /// Returns a vector containing the BSON documents that were read.
-    pub fn next_batch(&mut self) -> Result<Vec<bson::Document>> {
-        let n = self.batch_size;
-        self.next_n(n)
+    pub fn get_current_batch(&mut self) -> Result<Vec<bson::Document>> {
```
saghm: Given that the default batch size is being changed zero, the implementation of this method needs to be changed as well. Since a batch size of 0 means "let the server decide the batch size", I don't think it makes sense to hard code a specific number of documents to fetch in this method, so the most straightforward implementation would be to simply return the remaining documents from the current batch. While this changes what the method will return in the case that the buffer is partially empty, I think this is more in line with my expectations of what this method should be doing.

The name `next_batch` doesn't quite seem to fit with the proposed new behavior of the method, so I also propose changing it to `get_current_batch`, which is a more accurate description.

src/cursor.rs
```diff
@@ -36,7 +36,7 @@ use wire_protocol::operations::Message;
 
 use std::collections::vec_deque::VecDeque;
 
-pub const DEFAULT_BATCH_SIZE: i32 = 20;
+pub const DEFAULT_BATCH_SIZE: i32 = 0;
```
kyeah: Let's add a little comment here about this allowing the server to decide the batch size?

src/cursor.rs
```diff
@@ -469,9 +469,12 @@ impl Cursor {
     /// # Return value
     ///
     /// Returns a vector containing the BSON documents that were read.
-    pub fn next_batch(&mut self) -> Result<Vec<bson::Document>> {
-        let n = self.batch_size;
-        self.next_n(n)
```
kyeah: let's use the deprecation feature here, and default to 20 as before? we can suggest clients use `cursor.next_n(cursor.batch_size)` if they need to replicate the exact functionality.

src/cursor.rs
kyeah: `drain_or_get_batch` might be more descriptive here!

saghm: It feels somewhat ambiguous whether the grouping of that name is "[drain or get] batch" or "[drain] or [get next batch]". I know that the latter is somewhat nonsensical in this context, but it still feels like an awkward name that requires some logical overhead. Maybe something like `drain_current_batch`?

src/cursor.rs
saghm: Sounds good.

src/cursor.rs
saghm: Fair enough, although I'd be in favor of getting rid of this at some point in the future.

src/cursor.rs
kyeah: 👍 SGTM.

Issue #191, opened 2017-02-24
Investigate r2d2 for connection poolilng


kyeah:
We should investigate [r2d2](https://github.com/sfackler/r2d2) for connection pooling and see if that gives us any performance benefits, and if so, whether there are any downsides to switching.

saghm:
Without having had to chance to actually look into the library, the name of it alone already puts it at an advantage in my mind.

mexus:
Hi guys!

Any plans on implementing that feature? It might be very useful for real-world applications. Could it be that you lack on resources and some help might be used? :)

saghm:
Hello everyone! We've just released an alpha of [an official MongoDB Rust driver](https://github.com/mongodb/mongo-rust-driver); you can read about it [here](https://www.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver)! Because of this, we won't be updating this driver anymore, and the repository will be made read-only as soon as I get finished posting these messages. Thanks for using this driver, and we hope you try out the new one!

PR #218, opened 2017-05-28
refactor ClientOptions into Client builder

saghm:
This is the first step towards #191. Since ClientOptions seems like the most reasonable choice of what to implement [`ManageConnection`](https://docs.rs/r2d2/0.7.2/r2d2/trait.ManageConnection.html)  on, I've refactored it into a builder for Clients, which now no longer have any `connect` methods on them.

One thing that we've run into in regards to connections pools on the C++ driver is that changing the state of a client (e.g. setting a read preference or write concern) and then returning it to the pool means that the next time a user obtains that client from the pool, it will already have the custom state, which is generally not expected. We came to the conclusion that since setting the read preference or write concern at the client level rather than the database or collection level (or just for a single operation) is generally not a good way of doing things in the first place (as it only exists for legacy reasons anyhow), we would just document that setting the read preference, etc. on a client from a pool is a programming error. Given that this is already making a large number of breaking changes, I figured it would be reasonable to remove those options from being present at the client level, since we'd also run into the same issue once we move to a model where we get clients from a pool.

As always, I ran rustfmt, which came with its own set of changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,12 @@ Here's a basic example of driver usage:
 
 ```rust
 use bson::Bson;
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
 use mongodb::db::ThreadedDatabase;
 
 fn main() {
-    let client = Client::connect("localhost", 27017)
+    let connnector = Connector::new();
+    let client = connector.connect("localhost", 27017)
         .expect("Failed to initialize standalone client.");
 
     let coll = client.db("test").collection("movies");
@@ -83,7 +84,7 @@ fn main() {
 }
 ```
 
-To connect with SSL, use `Connector::with_ssl` and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
+To connect with SSL, use `Connector::ssl`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
 
 ```rust
 use bson::Bson;
@@ -100,9 +101,9 @@ fn main() {
     // Whether or not to verify that the server certificate is valid. Unless you're just testing out something locally, this should ALWAYS be true.
     let verify_peer = true;
 
-    let options = Connector::with_ssl(ca_file, certificate, key_file, verify_peer);
+    let options = Connector::ssl(ca_file, certificate, key_file, verify_peer);
 
-    let client = Client::connect_with_options("localhost", 27017, options)
+    let client = connector.connect("localhost", 27017)
         .expect("Failed to initialize standalone client.");
 
     // Insert document into 'test.movies' collection
--- a/README.md
+++ b/README.md
@@ -83,11 +83,11 @@ fn main() {
 }
 ```
 
-To connect with SSL, use `ClientOptions::with_ssl` and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
+To connect with SSL, use `Connector::with_ssl` and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
 
 ```rust
 use bson::Bson;
-use mongodb::{Client, ClientOptions, ThreadedClient};
+use mongodb::{Client, Connector, ThreadedClient};
 use mongodb::db::ThreadedDatabase;
 
 fn main() {
@@ -100,7 +100,7 @@ fn main() {
     // Whether or not to verify that the server certificate is valid. Unless you're just testing out something locally, this should ALWAYS be true.
     let verify_peer = true;
 
-    let options = ClientOptions::with_ssl(ca_file, certificate, key_file, verify_peer);
+    let options = Connector::with_ssl(ca_file, certificate, key_file, verify_peer);
 
     let client = Client::connect_with_options("localhost", 27017, options)
         .expect("Failed to initialize standalone client.");
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ fn main() {
 }
 ```
 
-To connect with SSL, use `Connector::ssl`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
+To connect with SSL, use `Connector::connect_with_ssl`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
 
 ```rust
 use bson::Bson;
@@ -101,9 +101,13 @@ fn main() {
     // Whether or not to verify that the server certificate is valid. Unless you're just testing out something locally, this should ALWAYS be true.
     let verify_peer = true;
 
-    let options = Connector::ssl(ca_file, certificate, key_file, verify_peer);
+    let options = Connector::ssl();
 
-    let client = connector.connect("localhost", 27017)
+    let client = connector.connect_with_ssl("mongodb://localhost:27017", 
+                                            ca_file,
+                                            certificate,
+                                            key_file,
+                                            verify_peer)
         .expect("Failed to initialize standalone client.");
 
     // Insert document into 'test.movies' collection
--- a/src/connstring.rs
+++ b/src/connstring.rs
@@ -95,85 +95,91 @@ impl ConnectionString {
             options: None,
         }
     }
+
+    /// Parses a MongoDB connection string URI as defined by
+    /// [the manual](http://docs.mongodb.org/manual/reference/connection-string/).
+    pub fn parse(address: &str) -> Result<ConnectionString> {
+        if !address.starts_with(URI_SCHEME) {
+            return Err(ArgumentError(String::from("MongoDB connection string must start with \
+                                               'mongodb://'.")));
+        }
+
+        // Remove scheme
+        let addr = &address[URI_SCHEME.len()..];
+
+        let hosts: Vec<Host>;
+        let mut user: Option<String> = None;
+        let mut password: Option<String> = None;
+        let mut database: Option<String> = Some(String::from("test"));
+        let mut collection: Option<String> = None;
+        let mut options: Option<ConnectionOptions> = None;
+
+        // Split on host/path
+        let (host_str, path_str) = if addr.contains(".sock") {
+            // Partition ipc socket
+            let (host_part, path_part) = rsplit(addr, ".sock");
+            if path_part.starts_with('/') {
+                (host_part, &path_part[1..])
+            } else {
+                (host_part, path_part)
+            }
+        } else {
+            // Partition standard format
+            partition(addr, "/")
+        };
+
+        if path_str.is_empty() && host_str.contains('?') {
+            return Err(ArgumentError(String::from("A '/' is required between the host list and any \
+                                               options.")));
+        }
+
+        // Split on authentication and hosts
+        if host_str.contains('@') {
+            let (user_info, host_string) = rpartition(host_str, "@");
+            let (u, p) = try!(parse_user_info(user_info));
+            user = Some(String::from(u));
+            password = Some(String::from(p));
+            hosts = try!(split_hosts(host_string));
+        } else {
+            hosts = try!(split_hosts(host_str));
+        }
+
+        let mut opts = "";
+
+        // Split on database name, collection, and options
+        if !path_str.is_empty() {
+            if path_str.starts_with('?') {
+                opts = &path_str[1..];
+            } else {
+                let (dbase, options) = partition(path_str, "?");
+                let (dbase_new, coll) = partition(dbase, ".");
+                database = Some(String::from(dbase_new));
+                collection = Some(String::from(coll));
+                opts = options;
+            }
+        }
+
+        // Collect options if any exist
+        if !opts.is_empty() {
+            options = Some(split_options(opts).unwrap());
+        }
+
+        Ok(ConnectionString {
+               hosts: hosts,
+               string: Some(String::from(address)),
+               user: user,
+               password: password,
+               database: database,
+               collection: collection,
+               options: options,
+           })
+    }
 }
 
-/// Parses a MongoDB connection string URI as defined by
-/// [the manual](http://docs.mongodb.org/manual/reference/connection-string/).
-pub fn parse(address: &str) -> Result<ConnectionString> {
-    if !address.starts_with(URI_SCHEME) {
-        return Err(ArgumentError(String::from("MongoDB connection string must start with \
-                                               'mongodb://'.")));
+impl Default for ConnectionString {
+    fn default() -> Self {
+        Self::new("localhost", 27017)
     }
-
-    // Remove scheme
-    let addr = &address[URI_SCHEME.len()..];
-
-    let hosts: Vec<Host>;
-    let mut user: Option<String> = None;
-    let mut password: Option<String> = None;
-    let mut database: Option<String> = Some(String::from("test"));
-    let mut collection: Option<String> = None;
-    let mut options: Option<ConnectionOptions> = None;
-
-    // Split on host/path
-    let (host_str, path_str) = if addr.contains(".sock") {
-        // Partition ipc socket
-        let (host_part, path_part) = rsplit(addr, ".sock");
-        if path_part.starts_with('/') {
-            (host_part, &path_part[1..])
-        } else {
-            (host_part, path_part)
-        }
-    } else {
-        // Partition standard format
-        partition(addr, "/")
-    };
-
-    if path_str.is_empty() && host_str.contains('?') {
-        return Err(ArgumentError(String::from("A '/' is required between the host list and any \
-                                               options.")));
-    }
-
-    // Split on authentication and hosts
-    if host_str.contains('@') {
-        let (user_info, host_string) = rpartition(host_str, "@");
-        let (u, p) = try!(parse_user_info(user_info));
-        user = Some(String::from(u));
-        password = Some(String::from(p));
-        hosts = try!(split_hosts(host_string));
-    } else {
-        hosts = try!(split_hosts(host_str));
-    }
-
-    let mut opts = "";
-
-    // Split on database name, collection, and options
-    if !path_str.is_empty() {
-        if path_str.starts_with('?') {
-            opts = &path_str[1..];
-        } else {
-            let (dbase, options) = partition(path_str, "?");
-            let (dbase_new, coll) = partition(dbase, ".");
-            database = Some(String::from(dbase_new));
-            collection = Some(String::from(coll));
-            opts = options;
-        }
-    }
-
-    // Collect options if any exist
-    if !opts.is_empty() {
-        options = Some(split_options(opts).unwrap());
-    }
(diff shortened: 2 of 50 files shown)


README.md
```diff
@@ -49,11 +49,12 @@ Here's a basic example of driver usage:
 
 ```rust
 use bson::Bson;
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Updated README to use new API

src/connstring.rs
```diff
@@ -95,85 +95,91 @@ impl ConnectionString {
...
-}
 
-/// Parses a MongoDB connection string URI as defined by
-/// [the manual](http://docs.mongodb.org/manual/reference/connection-string/).
-pub fn parse(address: &str) -> Result<ConnectionString> {
-    if !address.starts_with(URI_SCHEME) {
-        return Err(ArgumentError(String::from("MongoDB connection string must start with \
+    /// Parses a MongoDB connection string URI as defined by
+    /// [the manual](http://docs.mongodb.org/manual/reference/connection-string/).
+    pub fn parse(address: &str) -> Result<ConnectionString> {
```
saghm: Since I was in the process of making breaking changes, I thought it made sense to move this to  a non-instance method on ConnectionString rather than a free method.

src/connstring.rs
```diff
@@ -95,85 +95,91 @@ impl ConnectionString {
...
-    Ok(ConnectionString {
-        hosts: hosts,
-        string: Some(String::from(address)),
-        user: user,
-        password: password,
-        database: database,
-        collection: collection,
-        options: options,
-    })
+impl Default for ConnectionString {
```
saghm: This seemed like a reasonable thing to add in.

src/cursor.rs
```diff
@@ -4,12 +4,13 @@
 //! # #[macro_use] extern crate bson;
 //! # extern crate mongodb;
 //! #
-//! # use mongodb::{Client, ThreadedClient};
+//! # use mongodb::{Connector, ThreadedClient};
```
saghm: This example was updates so that the test would pass. All other changes to the file are from rustfmt.

src/db/mod.rs
```diff
@@ -8,9 +8,11 @@
 //! ## Collection Operations
 //!
 //! ```no_run
-//! # use mongodb::{Client, ThreadedClient};
+//! # use mongodb::{Connector, ThreadedClient};
```
saghm: These examples were updated so that the tests would pass.

src/db/mod.rs
```diff
@@ -169,15 +175,14 @@ impl ThreadedDatabase for Database {
...
-        let rp = read_preference.unwrap_or_else(|| client.read_preference.to_owned());
-        let wc = write_concern.unwrap_or_else(|| client.write_concern.to_owned());
-
         Arc::new(DatabaseInner {
-            name: String::from(name),
-            client: client,
-            read_preference: rp,
-            write_concern: wc,
-        })
+                     name: String::from(name),
```
saghm: Because I removed the read preference and write concern options from `Client`, I changed the defaults for them on `Database` to be the previous defaults that `Client` had.

src/gridfs/file.rs
```diff
@@ -618,7 +622,7 @@ impl GfsFile {
             "chunkSize" => (self.chunk_size),
             "length" => (self.len),
             "md5" => (self.md5.to_owned()),
-            "uploadDate" => (self.upload_date.as_ref().unwrap().clone())
+            "uploadDate" => (self.upload_date.unwrap())
```
saghm: Clippy noticed that the date type implements `Copy`, so rather than taking a reference and cloning it, I updated it to implicitly copy.

src/gridfs/mod.rs
```diff
@@ -14,10 +14,10 @@
 //! which you want access without having to load the entire file into memory.
 //!
 //! ```no_run
-//! # use mongodb::{Client, ThreadedClient};
+//! # use mongodb::{Connector, ThreadedClient};
```
saghm: This example was updated so that the tests would pass. All other changes to this file are from rustfmt.

src/pool.rs
```diff
@@ -3,7 +3,7 @@ use error::Error::{self, ArgumentError, OperationError};
 use error::Result;
 
 use connstring::Host;
-use stream::{Stream, StreamConnector};
+use stream::{Stream, ConnectMethod};
```
saghm: These changes are a mix of updating to use the new name of `StreamConnector` (i.e. `ConnectMethod`) and rustfmt changes.

src/stream.rs
```diff
@@ -11,7 +11,7 @@ use openssl::x509::X509_FILETYPE_PEM;
 
 /// Encapsulates the functionality for how to connect to the server.
 #[derive(Clone)]
-pub enum StreamConnector {
+pub enum ConnectMethod {
```
saghm: Because `ClientOptions` was getting renamed to `Connector`, I felt that this type required a change as well in order to be less confusing. I'm not completely convinced that `Connector` and `ConnectMethod` are the best names we can come up with though, so suggestions are welcome.

Additional changes beyond the renaming are from rustfmt.

src/topology/mod.rs
```diff
@@ -10,7 +10,7 @@ use bson::oid;
 use common::{ReadPreference, ReadMode};
 use connstring::{ConnectionString, Host};
 use pool::PooledStream;
-use stream::StreamConnector;
+use stream::ConnectMethod;
```
saghm: All changes in this file are from the rename of `StreamConnector`.

src/topology/monitor.rs
```diff
@@ -10,7 +10,7 @@ use command_type::CommandType;
 use connstring::{self, Host};
 use cursor::Cursor;
 use pool::ConnectionPool;
-use stream::StreamConnector;
+use stream::ConnectMethod;
```
saghm: Changes in this file are a mix of the renaming of `StreamConnector` and from rustfmt.

src/topology/server.rs
```diff
@@ -5,7 +5,7 @@ use Error::{self, OperationError};
 use bson::oid;
 use connstring::Host;
 use pool::{ConnectionPool, PooledStream};
-use stream::StreamConnector;
+use stream::ConnectMethod;
```
saghm: Changes to this file are a mix of the renaming of `StreamConnector` and from rustfmt.

tests/apm/mod.rs
```diff
@@ -2,15 +2,17 @@ use std::fs::{self, File};
 use std::io::{BufRead, BufReader};
 
 use bson::Bson;
-use mongodb::{Client, ClientOptions, CommandResult, ThreadedClient};
+use mongodb::{Client, Connector, CommandResult, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/auth/mod.rs
```diff
@@ -1,36 +1,41 @@
 use bson::Bson;
-use mongodb::{CommandType, Client, ThreadedClient};
+use mongodb::{CommandType, Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/bulk.rs
```diff
@@ -1,24 +1,24 @@
 use bson::Bson;
 use mongodb::coll::options::WriteModel;
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/connstring.rs
```diff
@@ -1,87 +1,81 @@
-use mongodb::connstring;
+use mongodb::connstring::{self, ConnectionString};
```
saghm: Changes to this file are a mix of the moving of `connstring::parse` to `ConnectionString::parse` and from rustfmt.

tests/client/crud_spec/framework.rs
```diff
@@ -245,7 +245,7 @@ macro_rules! run_suite {
     ( $file:expr, $coll:expr ) => {{
         let json = Value::from_file($file).unwrap();
         let suite = json.get_suite().unwrap();
-        let client =  Client::connect("localhost", 27017).unwrap();
+        let client =  Connector::new().connect("localhost", 27017).unwrap();
```
saghm: This change is to use the updated API.

tests/client/crud_spec/read.rs
```diff
@@ -2,7 +2,7 @@ use bson::Bson;
 use json::crud::arguments::Arguments;
 use json::crud::reader::SuiteContainer;
 use json::eq::{self, NumEq};
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: This change is to use the updated API.

tests/client/crud_spec/write.rs
```diff
@@ -2,7 +2,7 @@ use bson::Bson;
 use json::crud::arguments::Arguments;
 use json::crud::reader::SuiteContainer;
 use json::eq::{self, NumEq};
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: This change is to use the updated API.

tests/client/cursor.rs
```diff
@@ -1,6 +1,6 @@
 use bson::{Bson, Document};
 
-use mongodb::{Client, CommandType, ThreadedClient};
+use mongodb::{CommandType, Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/db.rs
```diff
@@ -1,12 +1,12 @@
 use bson::{self, Bson};
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/mod.rs
```diff
@@ -9,24 +9,30 @@ mod gridfs;
 mod wire_protocol;
 
 use bson;
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/wire_protocol.rs
```diff
@@ -1,16 +1,16 @@
 use bson::{Bson, Document};
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/sdam/framework.rs
```diff
@@ -1,8 +1,8 @@
-use mongodb::{Client, ThreadedClient};
+use mongodb::Connector;
 use mongodb::Error::OperationError;
-use mongodb::connstring::{self, ConnectionString};
+use mongodb::connstring::ConnectionString;
```
saghm: Changes to this file are a mix of using the updated API, the moving of `connstring::parse` to `ConnectionString::parse`, and from rustfmt.

tests/sdam/single.rs
```diff
@@ -1,4 +1,4 @@
-use mongodb::stream::StreamConnector;
+use mongodb::stream::ConnectMethod;
```
saghm: These changes are from the renaming of `StreamConnector`.

tests/server_selection/framework.rs
```diff
@@ -1,7 +1,6 @@
-use mongodb::{Client, ThreadedClient};
+use mongodb::Connector;
```
saghm: Changes to this file are a mix of using the updated API, the renaming of `StreamConnector`, and from rustfmt.

tests/ssl/mod.rs
```diff
@@ -1,6 +1,6 @@
 use std::path::PathBuf;
 
-use mongodb::{Client, ClientOptions, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

tests/client/coll.rs
```diff
@@ -1,13 +1,13 @@
 use bson::Bson;
 
-use mongodb::{Client, ThreadedClient};
+use mongodb::{Connector, ThreadedClient};
```
saghm: Changes to this file are a mix of using the updated API and from rustfmt.

src/db/roles.rs
```diff
@@ -86,9 +86,4 @@ impl Role {
     pub fn to_bson(&self) -> Bson {
         self.clone().into()
     }
-
```
saghm: Given that clippy is unhappy about the name of this metthod and that it's deprecated, I removed it since there are already breaking changes in this PR.

src/lib.rs
```diff
@@ -7,22 +7,19 @@
 //! The Client is an entry-point to interacting with a MongoDB instance.
 //!
 //! ```no_run
-//! # use mongodb::{Client, ClientOptions, ThreadedClient};
-//! # use mongodb::common::{ReadMode, ReadPreference};
+//! # use mongodb::{Connector, ThreadedClient};
```
saghm: These examples are updated to pass the tests.

src/lib.rs
```diff
@@ -171,108 +170,147 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
-use common::{ReadPreference, ReadMode, WriteConcern};
+use common::{ReadPreference, WriteConcern};
 use connstring::ConnectionString;
 use db::{Database, ThreadedDatabase};
 use error::Error::ResponseError;
 use pool::PooledStream;
-use stream::StreamConnector;
-use topology::{Topology, TopologyDescription, TopologyType, DEFAULT_HEARTBEAT_FREQUENCY_MS,
-DEFAULT_LOCAL_THRESHOLD_MS, DEFAULT_SERVER_SELECTION_TIMEOUT_MS};
+use stream::ConnectMethod;
```
saghm: This change is from the renaming of `StreamConnector`.

src/lib.rs
```diff
@@ -84,17 +84,16 @@
 //!
 //! ## Connection Pooling
 //!
-//! Each server within a MongoDB server set is maintained by the driver with a separate connection
-//! pool. By default, each pool has a maximum of 5 concurrent open connections.
+//! XXX:Rewrite documentation here once r2d2 is integrated.
```
saghm: We'll need to rewrite the documentation for pools once we fully integrate r2d2.

src/lib.rs
```diff
@@ -171,108 +170,146 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
-use topology::{Topology, TopologyDescription, TopologyType, DEFAULT_HEARTBEAT_FREQUENCY_MS,
-DEFAULT_LOCAL_THRESHOLD_MS, DEFAULT_SERVER_SELECTION_TIMEOUT_MS};
+use stream::ConnectMethod;
+use topology::{Topology, TopologyDescription, DEFAULT_HEARTBEAT_FREQUENCY_MS,
+               DEFAULT_LOCAL_THRESHOLD_MS, DEFAULT_SERVER_SELECTION_TIMEOUT_MS};
 use topology::server::Server;
 
 /// Interfaces with a MongoDB server or replica set.
 pub struct ClientInner {
-    /// Indicates how a server should be selected for read operations.
```
saghm: As mentioned in the summary, I removed the read preference and write concern options from the client.

src/lib.rs
```diff
@@ -171,108 +170,146 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
     req_id: Arc<AtomicIsize>,
     topology: Topology,
     listener: Listener,
     log_file: Option<Mutex<File>>,
 }
 
 /// Configuration options for a client.
-#[derive(Default)]
-pub struct ClientOptions {
+pub struct Connector {
```
saghm: Given that `ClientOptions` is being converted to a builder for clients, I felt that it should be named better. My quick and dirty best approximation for how I view it is `Connector`, but there likely is a better way to describe it (maybe something relating to the r2d2 trait it will implement, `ManageConnection`).

src/lib.rs
```diff
@@ -171,108 +170,146 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
+                            key_file: &str,
+                            verify_peer: bool)
+                            -> Result<Client> {
+        let connect_method =
+            ConnectMethod::with_ssl(ca_file, certificate_file, key_file, verify_peer);
+        self.connect_with_connection_string(ConnectionString::parse(uri)?, connect_method)
+    }
+
+
+    fn connect_with_connection_string(&self,
```
saghm: This is essentially a copy of the old `Client::connect_with_config` except with the removal of read preference/write concern and the changing of variable names to reflect where they're now stored.

src/lib.rs
```diff
@@ -171,108 +170,146 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
     pub local_threshold_ms: i64,
-    /// Options for how to connect to the server.
-    pub stream_connector: StreamConnector,
 }
 
-impl ClientOptions {
-    /// Creates a new default options struct.
-    pub fn new() -> ClientOptions {
-        ClientOptions {
+impl Default for Connector {
```
saghm: I realized that deriving default on ClientOptions was a bad idea, as the numerical arguments should not default to 0.

src/lib.rs
```diff
@@ -171,108 +170,146 @@ use std::sync::atomic::{AtomicIsize, Ordering, ATOMIC_ISIZE_INIT};
...
-            write_concern: None,
             heartbeat_frequency_ms: DEFAULT_HEARTBEAT_FREQUENCY_MS,
             server_selection_timeout_ms: DEFAULT_SERVER_SELECTION_TIMEOUT_MS,
             local_threshold_ms: DEFAULT_LOCAL_THRESHOLD_MS,
-            stream_connector: StreamConnector::default(),
         }
     }
+}
+
+impl Connector {
```
saghm: This is the successor implementation of `ClientOptions`, with a builder method for adding a log file and a number of non-consuming `connect` methods to build clients.

saghm (commented):
I've annotated all changes that aren't from rustfmt to save you some time; unless you want to dig through all the formatting changes for some reason, you can just look at each comment (and the code it refers to) rather than the entire diff.

kyeah:
the changes look fine from a first pass — however, I'm wondering what the r2d2 implementation is going to look like if Connector is going to be implementing `ManageConnection`, but also returning Clients. I would expect a Connector to return a single stream, with a `Client` still living at the top-level, and using a Connector internally.

I think the implementation will mostly be the same (topology/server logic moving to the Connector), but this seems to imply a larger change to make Clients a single "connection", unless I'm misinterpreting this PR.

saghm:
Interesting; I was thinking that we were looking to expose the connection pooling at the top level rather than keeping hidden under the covers, but given that the current abstraction has worked for us, I can see the merits of keeping it. For now, how about we keep this PR open as a reference, and I can work on making a separate PR with the changes as you suggest so we can compare them?

kyeah:
sounds good @saghm 🙏

saghm:

PR #225, opened 2017-07-04
fix race condition causing deadlock when acquiring locks

### LINKED ISSUE #216: Failing to connect to a replica set with one member down.
I'm seeing lock poison errors and panics when I'm trying to connect to a replica set of a sharded cluster that has a member down:

```
thread '<unnamed>' panicked at 'rwlock read lock would result in deadlock', src/libstd/sys/unix/rwlock.rs:59
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::sys::imp::rwlock::RWLock::read
   6: std::sys_common::rwlock::RWLock::read
   7: <std::sync::rwlock::RwLock<T>>::read
   8: mongodb::topology::TopologyDescription::check_if_has_primary
   9: mongodb::topology::TopologyDescription::update_private
  10: mongodb::topology::TopologyDescription::update
  11: mongodb::topology::monitor::Monitor::update_top_description
  12: mongodb::topology::monitor::Monitor::set_err
  13: mongodb::topology::monitor::Monitor::execute_update
  14: mongodb::topology::monitor::Monitor::run
  15: mongodb::topology::server::Server::new::{{closure}}
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', src/libcore/result.rs:859
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::result::unwrap_failed
   9: <core::result::Result<T, E>>::unwrap
  10: mongodb::topology::monitor::Monitor::update_top_description
  11: mongodb::topology::monitor::Monitor::update_with_is_master_cursor
  12: mongodb::topology::monitor::Monitor::execute_update
  13: mongodb::topology::monitor::Monitor::run
  14: mongodb::topology::server::Server::new::{{closure}}
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', src/libcore/result.rs:859
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::result::unwrap_failed
   9: <core::result::Result<T, E>>::unwrap
  10: mongodb::topology::monitor::Monitor::update_top_description
  11: mongodb::topology::monitor::Monitor::update_with_is_master_cursor
  12: mongodb::topology::monitor::Monitor::execute_update
  13: mongodb::topology::monitor::Monitor::run
  14: mongodb::topology::server::Server::new::{{closure}}
```

I've tried setting `ReadMode::Primary` as `ReadPreference` but still it panics.

For context, I'm writing a tool to analyse the data from the profiling collections on each shard (for a specific database), so I get a list of all the shards from a `mongos` or the config servers and then connect to each shard from separate thread.

In this particular case, one of the replica sets that consists of one arbiter, two healthy member and a third failed/unreachable member was giving me these errors which basically kills the whole thread for this shard.
When I removed the faulty member I can connect again without problems.

saghm:
This fixes #216. From my description of the bug in the issue:

"We have a few types dealing with the state of the topology wrapped up in RwLocks that are shared between all the monitoring threads and all the connection threads. When the monitoring threads check to make sure that we didn't lose connection to any of the servers, they acquire locks on those data structures to be able to inspect their state. Before the fix, we were passing the lock guards through to the functions that handled the case of a topology update (due to things like a lost connection to a server or a new primary being elected). When updating the topology, one of the things the driver does is check each server for stale information and update it in the case that it does; however, updating this information requires acquiring a lock on the data structure that contains the stale information. From what I can tell, the race condition occurs when the monitoring thread that detects the need for a topology update happens to be monitoring a server that the driver has stale information for, which then causes the driver to try to acquire a write lock on the same data structure whose lock guard is already being passed around, which causes deadlock. RwLock tries to detect when deadlock might occur and will panic if it does so, which is what caused the PoisonLock error that you were getting. Having the driver pass around the RwLock itself instead of the acquired guard on it (and then only acquiring the lock when it's needed in the topology update and making sure to release it once it's no longer needed) ensures that there won't be any attempt to acquire a lock that's already been acquired, which removes the deadlock."

tl;dr We were passing a lock guard through the topology update functions rather than the `RwLock`'d struct itself, and sometimes we'd try to acquire a write lock on the same thing, which is obviously Very Bad (tm).

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -80,12 +80,12 @@ impl FromStr for TopologyType {
     type Err = Error;
     fn from_str(s: &str) -> Result<Self> {
         Ok(match s {
-               "Single" => TopologyType::Single,
-               "ReplicaSetNoPrimary" => TopologyType::ReplicaSetNoPrimary,
-               "ReplicaSetWithPrimary" => TopologyType::ReplicaSetWithPrimary,
-               "Sharded" => TopologyType::Sharded,
-               _ => TopologyType::Unknown,
-           })
+            "Single" => TopologyType::Single,
+            "ReplicaSetNoPrimary" => TopologyType::ReplicaSetNoPrimary,
+            "ReplicaSetWithPrimary" => TopologyType::ReplicaSetWithPrimary,
+            "Sharded" => TopologyType::Sharded,
+            _ => TopologyType::Unknown,
+        })
     }
 }
 
@@ -146,7 +146,9 @@ impl TopologyDescription {
                 }
             }
         }
-        Err(OperationError(String::from("No servers available for the provided ReadPreference.")))
+        Err(OperationError(String::from(
+            "No servers available for the provided ReadPreference.",
+        )))
     }
 
     /// Returns a random server stream from the vector.
@@ -164,18 +166,22 @@ impl TopologyDescription {
             }
             servers.remove(index);
         }
-        Err(OperationError(String::from("No servers available for the provided ReadPreference.")))
+        Err(OperationError(String::from(
+            "No servers available for the provided ReadPreference.",
+        )))
     }
 
     /// Returns a server stream for read operations.
-    pub fn acquire_stream(&self,
-                          read_preference: &ReadPreference)
-                          -> Result<(PooledStream, bool, bool)> {
+    pub fn acquire_stream(
+        &self,
+        read_preference: &ReadPreference,
+    ) -> Result<(PooledStream, bool, bool)> {
         let (mut hosts, rand) = self.choose_hosts(read_preference)?;
 
         // Filter hosts by tagsets
         if self.topology_type != TopologyType::Sharded &&
-           self.topology_type != TopologyType::Single {
+            self.topology_type != TopologyType::Single
+        {
             self.filter_hosts(&mut hosts, read_preference);
         }
 
@@ -312,15 +318,16 @@ impl TopologyDescription {
                 // If no tags match but the replica set has a primary that is returnable with
                 // the given ReadMode, return that primary server.
                 if self.topology_type == TopologyType::ReplicaSetWithPrimary &&
-                   (read_preference.mode == ReadMode::Primary ||
-                    read_preference.mode == ReadMode::PrimaryPreferred) {
+                    (read_preference.mode == ReadMode::Primary ||
+                         read_preference.mode == ReadMode::PrimaryPreferred)
+                {
                     // Retain primaries.
                     hosts.retain(|host| if let Some(server) = self.servers.get(host) {
-                                     let description = server.description.read().unwrap();
-                                     description.server_type == ServerType::RSPrimary
-                                 } else {
-                                     false
-                                 });
+                        let description = server.description.read().unwrap();
+                        description.server_type == ServerType::RSPrimary
+                    } else {
+                        false
+                    });
                 } else {
                     // If no tags match and the above case does not occur,
                     // filter out all provided servers.
@@ -361,21 +368,20 @@ impl TopologyDescription {
         }
 
         // Find the shortest round-trip time.
-        let shortest_rtt = hosts
-            .iter()
-            .fold({
-                      // Initialize the value to the first server's round-trip-time, or i64::MAX.
-                      if let Some(server) = self.servers.get(&hosts[0]) {
-                          if let Ok(description) = server.description.read() {
-                              description.round_trip_time.unwrap_or(i64::MAX)
-                          } else {
-                              i64::MAX
-                          }
-                      } else {
-                          i64::MAX
-                      }
-                  },
-                  |acc, host| {
+        let shortest_rtt = hosts.iter().fold(
+            {
+                // Initialize the value to the first server's round-trip-time, or i64::MAX.
+                if let Some(server) = self.servers.get(&hosts[0]) {
+                    if let Ok(description) = server.description.read() {
+                        description.round_trip_time.unwrap_or(i64::MAX)
+                    } else {
+                        i64::MAX
+                    }
+                } else {
+                    i64::MAX
+                }
+            },
+            |acc, host| {
                 // Compare the previous shortest rtt with the host rtt.
                 if let Some(server) = self.servers.get(host) {
                     if let Ok(description) = server.description.read() {
@@ -388,7 +394,8 @@ impl TopologyDescription {
                     }
                 }
                 acc
-            });
+            },
+        );
 
         // If the shortest rtt is i64::MAX, all server rtts are None or could not be read.
         if shortest_rtt == i64::MAX {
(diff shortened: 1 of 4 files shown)


src/topology/mod.rs
```diff
@@ -511,32 +520,38 @@ impl TopologyDescription {
...
     /// Update the topology description, but don't start any monitors for new servers.
-    pub fn update_without_monitor(&mut self,
-                                  host: Host,
-                                  description: ServerDescription,
-                                  client: Client,
-                                  top_arc: Arc<RwLock<TopologyDescription>>) {
+    pub fn update_without_monitor(
+        &mut self,
+        host: Host,
+        description: Arc<RwLock<ServerDescription>>,
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/mod.rs
```diff
@@ -511,32 +520,38 @@ impl TopologyDescription {
...
     /// Updates the topology description based on an updated server description.
-    pub fn update(&mut self,
-                  host: Host,
-                  description: ServerDescription,
-                  client: Client,
-                  top_arc: Arc<RwLock<TopologyDescription>>) {
+    pub fn update(
+        &mut self,
+        host: Host,
+        description: Arc<RwLock<ServerDescription>>,
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/mod.rs
```diff
@@ -511,32 +520,38 @@ impl TopologyDescription {
...
-                      description: ServerDescription,
-                      client: Client,
-                      top_arc: Arc<RwLock<TopologyDescription>>,
-                      run_monitor: bool) {
-
-        let stype = description.server_type;
+    fn update_private(
+        &mut self,
+        host: Host,
+        description: Arc<RwLock<ServerDescription>>,
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/mod.rs
```diff
@@ -628,53 +647,70 @@ impl TopologyDescription {
...
-    fn update_rs_from_primary(&mut self,
-                              host: Host,
-                              description: ServerDescription,
-                              client: Client,
-                              top_arc: Arc<RwLock<TopologyDescription>>,
-                              run_monitor: bool) {
+    fn update_rs_from_primary(
+        &mut self,
+        host: Host,
+        description: Arc<RwLock<ServerDescription>>,
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/mod.rs
```diff
@@ -628,53 +647,70 @@ impl TopologyDescription {
...
+                if let Some(max_set_version) = self.max_set_version {
+                    if let Some(max_election_id) = self.max_election_id.clone() {
+                        if max_set_version > set_version ||
+                            (max_set_version == set_version && max_election_id > election_id)
+                        {
+                            // Stale primary
+                            if let Some(server) = self.servers.get(&host) {
+                                {
+                                    let mut server_description =
+                                        server.description.write().unwrap();
```
saghm: I belive this is where the error occured; since we were already passing a lock guard for a `ServerDescription` to this point, if the server we selected here is the same one, then trying to acquire a write lock causes deadlock.

src/topology/mod.rs
```diff
@@ -759,41 +811,49 @@ impl TopologyDescription {
...
-    fn add_missing_hosts(&mut self,
-                         description: &ServerDescription,
-                         client: Client,
-                         top_arc: Arc<RwLock<TopologyDescription>>,
-                         run_monitor: bool) {
-
-        for host in &description.hosts {
+    fn add_missing_hosts(
+        &mut self,
+        description: Arc<RwLock<ServerDescription>>,
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/monitor.rs
```diff
@@ -298,25 +301,27 @@ impl Monitor {
         &self,
         doc: bson::Document,
         round_trip_time: i64,
-    ) -> Result<ServerDescription> {
+    ) -> Result<Arc<RwLock<ServerDescription>>> {
```
saghm: Changed this to return the locked description rather than just a guard

src/topology/monitor.rs
```diff
@@ -298,25 +301,27 @@ impl Monitor {
...
             }
         }
 
-        Ok(server_description.clone())
+        Ok(self.server_description.clone())
     }
 
     // Updates the topology description associated with this monitor using a new server description.
-    fn update_top_description(&self, description: ServerDescription) {
+    fn update_top_description(&self, description: Arc<RwLock<ServerDescription>>) {
```
saghm: Changed this to take the locked description rather than just a guard

src/topology/mod.rs
```diff
@@ -759,41 +804,49 @@ impl TopologyDescription {
...
+                    top_arc.clone(),
+                    run_monitor,
+                    self.stream_connector.clone(),
+                );
                 self.servers.insert(host.clone(), server);
             }
         }
 
-        for host in &description.arbiters {
+        for host in &description.read().unwrap().arbiters {
```
kyeah: same here — wondering if we should clone these vecs instead of locking the description?

src/topology/mod.rs
```diff
@@ -689,13 +718,17 @@ impl TopologyDescription {
...
         let mut hosts_to_remove = Vec::new();
         for host in self.servers.keys() {
-            if !description.hosts.contains(host) && !description.passives.contains(host) &&
-               !description.arbiters.contains(host) {
+            let description_guard = description.read().unwrap();
+
+            if !description_guard.hosts.contains(host) &&
+                !description_guard.passives.contains(host) &&
+                !description_guard.arbiters.contains(host)
+            {
```
saghm: With this kind of thing, I tend to be of the opinion that the only way to really determine what performs better is to actually run benchmarks rather than abstractly reasoning about it. Unfortunately, we still haven't had a chance to implement any benchmarks for the driver (although I think we have an issue open for it), so it's probably best to leave the behavior as possible close to as what it was before, which would be acquiring the lock only once as you suggest. I'll implement your suggestion before merging.

src/topology/mod.rs
saghm: I'll implement your suggestion here for the same reasons as above

Issue #223, opened 2017-06-30
Cannot create text indexes


CircuitCoder:
When creating text indexes, I get an error saying: "Index model keys must map to i32."

https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/master/src/coll/options.rs#L368

saghm:
Hi! Can you provide some code that reproduces this?

CircuitCoder:
According to the Javascript driver, a text index can be created in the following way, as said in http://mongodb.github.io/node-mongodb-native/2.2/tutorials/create-indexes/#create-a-text-index:
```javascript
coll.createIndex({
  field: 'text'
}, function callback() { /* ... */ });
```

But the following rust code:
```rust
coll.create_index(doc!{
    "field" => "text",
});
```

returns the error "Index model keys must map to i32.", because it only accepts integers for model keys.

Also, hashed indexes are created in the similar way, so I assume they currently cannot be created with the rust driver as well.

CircuitCoder:
The C driver's implementation of index name generation is here:
https://github.com/mongodb/mongo-c-driver/blob/master/src/mongoc/mongoc-collection.c#L1029-L1043

It just accepts UTF8 values and concat them with the model key, like the way it deals with integers values.

So I guess it's safe to add a few match arms to allow string values like "text", "hashed" and so on.

CircuitCoder:
I've just created a PR for this one.

saghm:
Thanks! Sorry for not having a chance to get to it on my own quickly enough

saghm:
Fixed in #228

PR #228, opened 2017-07-19
Allowed IndexModel keys to map to text, hashed, 2d and 2dsphere

CircuitCoder:
Currently, text indexes cannot be created with this driver. This issues has been described in #223.

This PR does nothing more than adding a match arm to accept values other than i32 in the IndexModel.

### DIFF
--- a/src/coll/options.rs
+++ b/src/coll/options.rs
@@ -365,7 +365,11 @@ impl IndexModel {
             name.push('_');
             match *bson {
                 Bson::I32(ref i) => name.push_str(&format!("{}", i)),
-                _ => return Err(ArgumentError(String::from("Index model keys must map to i32."))),
+                Bson::String(ref s) if s == "text" || s == "hashed" || s == "2d" || s == "2dsphere"
+                    => name.push_str(s),
+                _ => return Err(ArgumentError(String::from(
+                            "Index model keys must map to i32, \"text\", \"hashed\", \"2d\" or \"2dsphere\""
+                            ))),
             }
         }
         Ok(name)
--- a/tests/client/coll.rs
+++ b/tests/client/coll.rs
@@ -689,6 +689,87 @@ fn create_list_drop_indexes() {
 }
 
 #[test]
+fn create_text_hashed_2d_2dsphere_index() {
+    let client = Client::connect("localhost", 27017).unwrap();
+    let db = client.db("test-client-coll");
+    let coll = db.collection("create_text_hashed_2d_2dsphere_index");
+
+    coll.create_index(doc!{
+        "a" => "text"
+    }, None).expect("Failed to create text indexes");
+
+    coll.create_index(doc!{
+        "b" => "hashed"
+    }, None).expect("Failed to create hashed indexes");
+
+    coll.create_index(doc!{
+        "c" => "2d"
+    }, None).expect("Failed to create 2d indexes");
+
+    coll.create_index(doc!{
+        "d" => "2dsphere"
+    }, None).expect("Failed to create 2dsphere indexes");
+}
+
+#[test]
+fn block_ill_formed_index_models() {
+    assert!(IndexModel::new(doc!{ "test" => 1.1 }, None).name().is_err());
+    assert!(IndexModel::new(doc!{ "test" => 1i64 }, None).name().is_err());
+    assert!(IndexModel::new(doc!{ "test" => "arbitrystr" }, None).name().is_err());
+}
+
+#[test]
+fn create_query_text_index() {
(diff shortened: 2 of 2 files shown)


kyeah:
cool, thank you! can we add these as constants i.e. something like:

```rust
pub mod IndexType {
  const TEXT: &'static str = "text";
  ...

  const TYPES: [&'static str; 4] = [TEXT, ...];
}
```

saghm:
@kyeah I agree with using constants instead of the string literals, but I'm curious to hear your reasoning for needing a `TYPES` array. Additionally, instead of making a separate module for just four string literal constants, what do you think of making a general `constants` module that contains all of our constants rather than having them spread out across different files? Right now, we seem to just be putting them in whichever file we need them first, but that leads to some possibly unexpected results, like `DEFAULT_MAX_BSON_OBJECT_SIZE` being in the `topology::monitor` module.

kyeah:
I think it's important to namespace constants to avoid constant soup and to keep the context of constants closer to their related modules (both from the development standpoint of having information be closer to the code they affect, and from a usage standpoint of having them stored under a more unified module path when they are being brought in). 

That said, I think that some of our constants live deeper than they should be, and could be moved up into a more root-level module. Having `DEFAULT_MAX_BSON_OBJECT_SIZE` in the monitoring module perhaps indicates we should have a module related to mongod-related settings, or that we should break out `isMaster` logic.

```rust
mongodb::mongod::DEFAULT_MAX_BSON_OBJECT_SIZE
```

Thinking about a `constants` module, I think that is an option to have on a more fine-grained, per-module basis e.g. `mongodb::coll::constants` — again, to avoid constant soup. However, from a user-perspective, it's another level of indirection that isn't really necessary, as those constants should just be associated with the containing module, e.g. `mongodb::coll`. 

I also doubt we'll have any modules that get bloated enough with constants that we'd want to break them out into a separate file.

```rust
use mongodb::coll::IndexType;
use mongodb::coll::options::IndexModel;
```

Seems pretty reasonable IMO — people generally look for constants having some context of what that constant relates to.

On the point about `IndexType::TYPES`, you can see in this PR how it'd be helpful:

```rust
// before
Bson::String(ref s) if s == "text" || s == "hashed" || s == "2d" || s == "2dsphere"

// after
Bson::String(ref s) if IndexType::TYPES.contains(s)
```

saghm:
I don't feel super strongly about this, so I'll defer to your opinion on it. Do you want to edit the PR to your liking and then merge it (or just pull the branch from the fork, add a commit on top, and then push it)? It's been almost a month since your initial request, so I think it's fine to just do it ourselves.

As for having an array of all the index types, why not just skip having the individual constants themselves and just put the string literals right in the container? I don't think we'd need to use them individually within the driver, so it seems simpler to make it so that we only have to add one thing rather than two if new keys need to be added.

kyeah:
I'll follow up on this if @CircuitCoder doesn't before me :)

re: individual constants, I assumed these were constants that people could use for creating indexes instead of magic values, but I'm not familiar with our index api.

saghm:
I'm not convinced it's worth making these constants public, to be honest. It's extremely unlikely that MongoDB (the server) will change the string constants it requires to make a text index, and if I were a user, I don't think I'd bother to import a constant just to put the word "text" or something into a document. While some of the other literals like "2dsphere" might not be totally obvious without looking them up, the overwhelming majority of users would never need to use them, and those that do will likely already be familiar with the API for geospatial indexes.

kyeah:
The philosophy behind replacing magic values with constants is not necessarily to define/document the api within the codebase (although it may help), but to prevent simple human errors like typos from passing compilation and surfacing at runtime. It doesn't prevent people from using bare strings so much as enable those who wish to avoid stringy references and the innocent mistakes that stem from them. 

And while it is extremely unlikely that a breaking change is introduced, using these constants gives people a stronger contract that they can be 100% confident in.

saghm:
Once again, I don't have a very strong opinion on this, so I'll defer to your judgment :)

Honestly, if I were writing this from scratch, I don't even think I'd bother doing any validation on the document that the user provided for the index; index creation is almost always a one-time operation that's done outside of application itself, so for the minority of users that choose to create them through a driver rather than the shell, the cost of a single round trip to have the server report back errors itself seems worth it to me to not have to maintain the validation code (which would need to be updated if new types of indexes are added to the database or if a breaking change in the server API did change for some reason). But given that we're already doing validation checks, I suppose it's worth continuing to do so.

kyeah:
Going to merge this and follow up with a constants PR. Thanks @CircuitCoder! 🙏

Issue #232, opened 2017-08-15
aggregate too slow?


Matrix-Zhang:
i use aggregate to query documents where should have 110 results, i use iter to map the results to some type, elapsed 5 secs, how to optimize？

saghm:
Hi! Can you provide an example to reproduce this (self-contained code to run, contents of database to run against, etc.)?

Matrix-Zhang:
```
extern crate bson;
extern crate mongodb;
extern crate serde;
#[macro_use]
extern crate serde_derive;
extern crate serde_json;

use std::time::SystemTime;

use bson::{Bson, to_bson};
use mongodb::{Client, ThreadedClient};
use mongodb::db::ThreadedDatabase;
use serde_json::Value;

#[derive(Debug, Deserialize, Serialize)]
struct Model {
    id: String,
    logs: Vec<i32>,
}

fn main() {
    let client = Client::connect("localhost", 27017).expect("Failed to initialize client.");
    let collection = client.db("demo").collection("models");

    let mut documents = vec![];

    for i in 0..100 {
        let model = Model {
            id: i.to_string(),
            logs: vec![1, 2, 3, 4, 5],
        };

        documents.push(to_bson(&model).unwrap().as_document().unwrap().to_owned());
    }

    collection.insert_many(documents, None).unwrap();

    let project =
        doc!{
        "$project" => {
            "id" => 1,
            "last_log" => {
                "$slice" => ["$logs", (-1)]
            }
        }
    };

    let mut values: Vec<Value> = vec![];

    let time = SystemTime::now();

    let cursor = collection.aggregate(vec![project], None).unwrap();

    for result in cursor {
        if let Ok(document) = result {
            values.push(Bson::from(document).into());
        }
    }

    println!(
        "values's count: {}, elapsed: {}s",
        values.len(),
        time.elapsed().unwrap().as_secs()
    );

    collection.drop().unwrap();
}
```
the output is :
values's count: 100, elapsed: 4s

saghm:
[...]

PR #236, opened 2017-09-10
fix default batch size for aggregate

saghm:
As reported in #232, it turns out our fix to the default batch size a little while back actually broke the default batch size for aggregations; aggregations specify the batch size as an option in the command document rather than a flag in the write protocol message, and apparently the server doesn't interpret "0" as meaning "use the server default batch size here", but instead as "use a batch size of 1, and don't even return anything in the first batch. Fixing this just requires not setting the batch size for aggregations when the user doesn't specify one.

As usual, this PR comes chock full of rustfmt changes. You can ignore all of the changes besides the one I commented on, or you can look at how rustfmt seems to have actually improved to the point where the changes are actually improvements rather than just arbitrary.

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -38,12 +38,13 @@ impl Collection {
     /// Creates a collection representation with optional read and write controls.
     ///
     /// If `create` is specified, the collection will be explicitly created in the database.
-    pub fn new(db: Database,
-               name: &str,
-               create: bool,
-               read_preference: Option<ReadPreference>,
-               write_concern: Option<WriteConcern>)
-               -> Collection {
+    pub fn new(
+        db: Database,
+        name: &str,
+        create: bool,
+        read_preference: Option<ReadPreference>,
+        write_concern: Option<WriteConcern>,
+    ) -> Collection {
 
         let rp = read_preference.unwrap_or_else(|| db.read_preference.to_owned());
         let wc = write_concern.unwrap_or_else(|| db.write_concern.to_owned());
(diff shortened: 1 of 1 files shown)


src/coll/mod.rs
```diff
@@ -116,19 +114,23 @@ impl Collection {
                 spec = merge_options(spec, aggregate_options);
             }
             None => {
-                let cursor = doc! { "batchSize" => (::cursor::DEFAULT_BATCH_SIZE) };
```
saghm: This is the only change that matters; the server does not treat an explicit batch size of 0 on aggregation commands as meaning "use the server's default batch size", hence its removal.

saghm:
Random semi-related question: how opposed are you to me just rustfmt'ing the entire repo so we don't get as many of these unrelated changes littered throughout our PRs? I've gotten the impression that it's gotten more stable lately, and it seems to be indenting things in a much more sane way nowadays. I'll still submit it as a PR for you to check out, of course :)

kyeah:
> Random semi-related question: how opposed are you to me just rustfmt'ing the entire repo so we don't get as many of these unrelated changes littered throughout our PRs? I've gotten the impression that it's gotten more stable lately, and it seems to be indenting things in a much more sane way nowadays. I'll still submit it as a PR for you to check out, of course :)

Cool with that. If we could try to run rustfmt into separate PRs as much as possible (so we don't have to do actual-changes hide and seek) that would be wonderful.

saghm:
Awesome! I'll make a PR now, and I'll try to make a point to make another PR whenever a new rustfmt release comes out.

Issue #235, opened 2017-09-05
Do not enforce client cert and key for `ClientOptions::with_ssl`


lukaspustina:
Dear crate team,

thanks for this helpful crate for which I'd like to propose a change.

The function `ClientOptions::with_ssl` enforces a crate user to supply a CA cert as well as a client cert and a corresponding key. This is unfortunate, because this requirement steams from the code and not from the actual mongo DB driver. For example, it's perfectly fine to create an SSL encrypted connection using only use a CA cert -- cf. https://docs.mongodb.com/manual/tutorial/configure-ssl-clients/, Connect to MongoDB Instance with SSL Encryption.

The requirement is encoded in https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/master/src/stream.rs#L81. It'd be very easy to mitigate this requirement by using `Option` for at least `certificate_file` and `key_file` and set the openssl ssl_context accordingly. Unfortunately, introducing `Option` would break the current API of `ClientOptions::with_ssl`. The other way around this is to check for the empty string `""` which is a anti-pattern.

Anyway, probably you guys have a better pictures of this and I'd like to start a discussion to evolve this. I'm happy to contribute code, once we agree on a design.

Cheers,
Lukas

saghm:
I think you might be misreading the documentation that you linked to. Specifically, the [`--sslPEMKeyFile`](https://docs.mongodb.com/manual/reference/program/mongostat/#cmdoption-sslpemkeyfile) option that the documentation describes requires that a client certificate is present in the file as well. This driver slightly diverges from that API by accepting the client certficate and client key as separate files instead of as a single file (although we plan to add the ability to use a single file as well), but the actual requirements are the same for the client credentials. (The requirements as a whole are slightly different in that the driver currently requires specifying a custom CA certificate  rather than defaulting to using the system CA store, but I don't think this is what you're asking about).

lukaspustina:
Thanks for your quick response. I think I didn't make myself clear.

The issue I like to resolve is that the current implementation requires to supply a CA certificate as well as client certificate. For a transport encrypted connection the client certificate is not necessary, but has to be supplied anyway though.

I'd like to change the crate in a way that only the CA certificate has to be passed to the factory method enabling TLS _without_ a client certificate. Without suggesting this change, but for demonstration purpose, a modified method signature could be:

```
[...]

PR #238, opened 2017-09-12
[#235] Add `with_unauthenticated_ssl` for unauthenticated ssl encryption

lukaspustina:
This commit allows to create an SSL encrypted connection without requiring to use client authentication -- cf. #235.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ fn main() {
 }
 ```
 
-To connect with SSL, use `ClientOptions::with_ssl`, `ClientOptions::with_unauthenticated_ssl`, and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
+To connect with SSL, use `ClientOptions::with_ssl` or `ClientOptions::with_unauthenticated_ssl`, and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
 
 ```rust
 use bson::Bson;
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ fn main() {
 }
 ```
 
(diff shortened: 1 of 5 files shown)


lukaspustina:
My bad. I'll check why the tests failed.

README.md
```diff
@@ -83,7 +83,7 @@ fn main() {
 }
 ```
 
-To connect with SSL, use `ClientOptions::with_ssl` and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
+To connect with SSL, use `ClientOptions::with_ssl`, `ClientOptions::with_unauthenticated_ssl`, and `Client::connect_with_options`. Afterwards, the client can be used as above (note that the server will have to be configured to accept SSL connections and that you'll have to generate your own keys and certificates):
```
saghm: I think this should read "`use  either`ClientOptions::with_ssl` or `ClientOptions::with_unauthenticated_ssl`, and then `Client::connect_with_options`.

src/stream.rs
```diff
@@ -18,8 +18,8 @@ pub enum StreamConnector {
     /// Connect to the server through a TCP stream encrypted with SSL.
     Ssl {
         ca_file: String,
-        certificate_file: String,
-        key_file: String,
+        certificate_file: Option<String>,
```
saghm: Given that this is public, I think it's worth adding another paragraph in the documentation for StreamConnector::Ssl noting that it's not valid to have one of the options populated but not the other.

lukaspustina: Done.

lukaspustina:
Anything else I can do?

saghm:
Everything looks good. Thanks so much for the PR! Assuming everything passes on Travis after the merge, I'll publish a new version with the changes.

lukaspustina:
Happy to contribute and thankful for your mongo driver! Cheers.

PR #239, opened 2017-10-03
Add a supplemental section on related projects.

thedodd:
Hello all. Awesome work with this driver!

Given how important it is to model data, I've created a project called [Wither](https://github.com/thedodd/wither) which is based on this library. It is a pretty minimal wrapper around the functionality that is already in this project, but allows you to reason about your data in a "model first" sort of fashion.

Seems like a reasonable idea to cross link the related projects to help drive adoption and feedback for both projects.

Any and all feedback is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,3 +150,8 @@ Run tests like a regular Rust program:
 ```
 cargo test --verbose
 ```
+
+Supplemental Projects
+---------------------
+
+- [Wither](https://github.com/thedodd/wither) An ODM for MongoDB built upon the mongo rust driver.


saghm:
Hi @thedodd! I've been casually following your project, and I think it's really interesting! I'd definitely be interested in working with you with regards to any support/changes you need from the driver, and my co-maintainer @kyeah has mentioned to me that he might be interested in contributing if you need any help.

Unfortunately, we don't think that adding a list of supplemental projects to the README is an option right now. Given that this repository is owned by MongoDB, we'd like to avoid confusion around what is or isn't officially supported by the company (which already has been a source of confusion with regards to the driver itself). As far as I'm aware, none of the other MongoDB-owned drivers reference supplemental projects in their README; the Ruby driver doesn't even mention mongoid (the Ruby MongoDB ODM) even though they're both maintained by the same person and mongoid is officially supported by the company.

I don't want this to discourage you in any way from continuing the work on Wither; I certainly think that having an ODM around the Rust driver is desirable given that the community is very fond of Diesel, and I'm looking forward to seeing the project mature and grow. Feel free to reach out either through Github or by emailing me directly (@ saghmrossi@gmail.com) if you need any help with using the Rust driver or if any changes in the driver would help the development of Wither!

thedodd:
+1

PR #241, opened 2017-10-15
Use new bson macro standard

kyeah:
## What

Update our usage of the bson macros to utilize colons + now-optional parens.

## Why

- [x] taking care of our house
- [x] encouraging this new, preferred style for users :)

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -191,7 +191,7 @@ impl Authenticator {
             doc! {
                 "saslContinue": 1,
                 "payload": binary,
-                "conversationId": (initial_data.conversation_id.clone())
+                "conversationId": initial_data.conversation_id.clone()
             };
 
         let response = try!(self.db.command(next_doc, Suppressed, None));
@@ -207,7 +207,7 @@ impl Authenticator {
         let final_doc =
             doc! {
                 "saslContinue": 1,
-                "payload": (Binary(Generic, vec![])),
+                "payload": Binary(Generic, vec![]),
                 "conversationId": conversation_id
             };
 
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -63,10 +63,10 @@ impl Authenticator {
 
         let start_doc =
             doc! {
-            "saslStart" => 1,
-            "autoAuthorize" => 1,
-            "payload" => binary,
-            "mechanism" => "SCRAM-SHA-1"
+            "saslStart": 1,
+            "autoAuthorize": 1,
+            "payload": binary,
+            "mechanism": "SCRAM-SHA-1"
         };
 
         let doc = try!(self.db.command(start_doc, Suppressed, None));
@@ -189,9 +189,9 @@ impl Authenticator {
 
         let next_doc =
             doc! {
-                "saslContinue" => 1,
-                "payload" => binary,
-                "conversationId" => (initial_data.conversation_id.clone())
+                "saslContinue": 1,
+                "payload": binary,
+                "conversationId": (initial_data.conversation_id.clone())
             };
 
         let response = try!(self.db.command(next_doc, Suppressed, None));
@@ -206,9 +206,9 @@ impl Authenticator {
     fn finish(&self, conversation_id: Bson, auth_data: AuthData) -> Result<()> {
         let final_doc =
             doc! {
-                "saslContinue" => 1,
-                "payload" => (Binary(Generic, vec![])),
-                "conversationId" => conversation_id
+                "saslContinue": 1,
+                "payload": (Binary(Generic, vec![])),
+                "conversationId": conversation_id
             };
 
         // Compute server key
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -99,8 +99,8 @@ impl Collection {
 
         let mut spec =
             doc! {
-            "aggregate" => (self.name()),
-            "pipeline" => pipeline_map
+            "aggregate": (self.name()),
+            "pipeline": pipeline_map
         };
 
         let mut read_preference = self.read_preference.clone();
@@ -215,8 +215,8 @@ impl Collection {
         let doc = match find_options.sort {
             Some(ref sort_opt) => {
                 doc! {
-                    "$query" => (filter.unwrap_or_else(bson::Document::new)),
-                    "$orderby" => (Bson::Document(sort_opt.clone()))
+                    "$query": (filter.unwrap_or_else(bson::Document::new)),
+                    "$orderby": (Bson::Document(sort_opt.clone()))
                 }
             }
             None => filter.unwrap_or_else(bson::Document::new),
@@ -281,8 +281,8 @@ impl Collection {
     ) -> Result<Option<bson::Document>> {
         let mut cmd =
             doc! {
-            "findAndModify" => (self.name()),
-            "query" => filter
+            "findAndModify": (self.name()),
+            "query": filter
         };
 
         cmd = merge_options(cmd, options);
@@ -310,7 +310,7 @@ impl Collection {
             None => (None, None),
         };
 
-        let mut options_doc = doc! { "remove" => true };
+        let mut options_doc = doc! { "remove": true };
 
         if let Some(find_one_and_delete_options) = options {
             options_doc = merge_options(options_doc, find_one_and_delete_options);
@@ -340,7 +340,7 @@ impl Collection {
             None => (None, None),
         };
 
-        let mut options_doc = doc! { "update" => replacement };
+        let mut options_doc = doc! { "update": replacement };
 
         if let Some(find_one_and_replace_options) = options {
             options_doc = merge_options(options_doc, find_one_and_replace_options);
@@ -370,7 +370,7 @@ impl Collection {
             None => (None, None),
         };
 
-        let mut options_doc = doc! { "update" => update };
+        let mut options_doc = doc! { "update": update };
 
         if let Some(find_one_and_update_options) = options {
             options_doc = merge_options(options_doc, find_one_and_update_options);
@@ -639,8 +639,8 @@ impl Collection {
 
         let mut cmd =
             doc! {
-            "insert" => (self.name()),
-            "documents" => converted_docs
+            "insert": (self.name()),
+            "documents": converted_docs
         };
 
         if let Some(insert_options) = options {
@@ -842,9 +842,9 @@ impl Collection {
 
         let cmd =
             doc! {
-            "update" => (self.name()),
-            "updates" => updates,
-            "writeConcern" => (wc.to_bson())
+            "update": (self.name()),
+            "updates": updates,
+            "writeConcern": (wc.to_bson())
         };
 
         let result = try!(self.db.command(cmd, cmd_type, None));
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -99,7 +99,7 @@ impl Collection {
 
         let mut spec =
             doc! {
-            "aggregate": (self.name()),
+            "aggregate": self.name(),
             "pipeline": pipeline_map
         };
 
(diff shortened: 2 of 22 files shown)


src/auth.rs
```diff
@@ -63,10 +63,10 @@ impl Authenticator {
 
         let start_doc =
             doc! {
-            "saslStart" => 1,
-            "autoAuthorize" => 1,
-            "payload" => binary,
-            "mechanism" => "SCRAM-SHA-1"
+            "saslStart": 1,
```
saghm: Ugh, rustfmt, why...

I _think_ rustfmt won't change it back if the indentation is fixed here, but since this seems like it will happen every time, I'm fine with leaving it if you don't want to bother.

src/coll/mod.rs
```diff
@@ -639,8 +639,8 @@ impl Collection {
 
         let mut cmd =
             doc! {
-            "insert" => (self.name()),
-            "documents" => converted_docs
+            "insert": self.name(),
```
saghm: rustfmt again...

src/coll/mod.rs
```diff
@@ -842,9 +842,9 @@ impl Collection {
 
         let cmd =
             doc! {
-            "update" => (self.name()),
-            "updates" => updates,
-            "writeConcern" => (wc.to_bson())
+            "update": self.name(),
```
saghm: and again

src/cursor.rs
```diff
@@ -279,7 +279,7 @@ impl Cursor {
     ) -> Result<Cursor> {
 
         let mut stream = stream;
-        let mut socket = stream.get_socket();
+        let socket = stream.get_socket();
```
saghm: This seems unrelated;  did the compiler give a warning here or something?

src/cursor.rs
```diff
@@ -298,8 +298,8 @@ impl Cursor {
             CommandType::Find => {
                 let document =
                     doc! { 
-                    "find" => coll_name,
-                    "filter" => filter
+                    "find": coll_name,
```
saghm: and again

src/cursor.rs
```diff
@@ -379,12 +379,12 @@ impl Cursor {
         let reply = match cmd_type {
             CommandType::Find => {
                 doc! {
-                "cursor" => {
-                    "id" => cursor_id,
-                    "ns" => (&namespace[..]),
-                    "firstBatch" => (Bson::Array(vec))
+                "cursor": {
```
saghm: and again

src/cursor.rs
```diff
@@ -418,7 +418,7 @@ impl Cursor {
 
     fn get_from_stream(&mut self) -> Result<()> {
         let (mut stream, _, _) = try!(self.client.acquire_stream(self.read_preference.to_owned()));
-        let mut socket = stream.get_socket();
+        let socket = stream.get_socket();
```
saghm: Ditto above about unrelated change

src/db/mod.rs
```diff
@@ -341,8 +341,8 @@ impl ThreadedDatabase for Database {
     ) -> Result<()> {
         let mut doc =
             doc! {
-            "createUser" => name,
-            "pwd" => password
+            "createUser": name,
```
saghm: and again

src/db/mod.rs
```diff
@@ -400,8 +400,8 @@ impl ThreadedDatabase for Database {
     fn get_all_users(&self, show_credentials: bool) -> Result<Vec<bson::Document>> {
         let doc =
             doc! {
-            "usersInfo" => 1,
-            "showCredentials" => show_credentials
+            "usersInfo": 1,
```
saghm: and again

src/db/mod.rs
```diff
@@ -425,7 +425,7 @@ impl ThreadedDatabase for Database {
     fn get_user(&self, user: &str, options: Option<UserInfoOptions>) -> Result<bson::Document> {
         let mut doc =
             doc! {
-            "usersInfo" => { "user" => user, "db" => (self.name.to_owned()) }
+            "usersInfo": { "user": user, "db": self.name.to_owned() }
```
saghm: For multiple keys, I think it's probably cleaner to have things on multiple lines. (rustfmt will still probably be weird about the indentation)

src/db/mod.rs
```diff
@@ -456,12 +456,12 @@ impl ThreadedDatabase for Database {
         let vec: Vec<_> = users
             .into_iter()
             .map(|user| {
-                let doc = doc! { "user" => user, "db" => (self.name.to_owned()) };
+                let doc = doc! { "user": user, "db": self.name.to_owned() };
```
saghm: Ditto above about multiple keys

src/gridfs/file.rs
```diff
@@ -201,17 +201,17 @@ impl File {
...
 
                 // Ensure indexes
-                try!(self.gfs.files.create_index(doc!{ "filename" => 1 }, None));
+                try!(self.gfs.files.create_index(doc!{ "filename": 1 }, None));
 
                 let mut opts = IndexOptions::new();
                 opts.unique = Some(true);
                 try!(self.gfs.chunks.create_index(
-                    doc!{ "files_id" => 1, "n" => 1},
+                    doc!{ "files_id": 1, "n": 1},
```
saghm: Ditto above about multiple keys

It's not super egregious here since they're small, but I think that it will never look worse to split it, and I think consistency is better than just winging it

src/gridfs/file.rs
```diff
@@ -246,10 +246,10 @@ impl File {
 
         let document =
             doc! {
-            "_id" => (try!(oid::ObjectId::new())),
-            "files_id" => (self.doc.id.clone()),
-            "n" => n,
-            "data" => (BinarySubtype::Generic, vec_buf)
+            "_id": try!(oid::ObjectId::new()),
```
saghm: Unrelated: do we want to do a mass switch from `try!` to `?` at some point? The only downside would be breaking things on older compilers, but I don't get the sense that Rust users generally stick to old stable versions

src/gridfs/file.rs
```diff
@@ -278,7 +278,7 @@ impl File {
 
     // Retrieves a binary file chunk from GridFS.
     pub fn find_chunk(&mut self, id: oid::ObjectId, chunk_num: i32) -> Result<Vec<u8>> {
-        let filter = doc!{"files_id" => id, "n" => chunk_num };
+        let filter = doc!{"files_id": id, "n": chunk_num };
```
saghm: Ditto above about multiple keys

src/gridfs/file.rs
```diff
@@ -328,7 +328,7 @@ impl File {
                 };
 
                 let result = arc_gfs.chunks.find_one(
-                    Some(doc!{"files_id" => (id), "n" => (next_chunk_num)}),
+                    Some(doc!{ "files_id": id, "n": next_chunk_num }),
```
saghm: Ditto above about multiple keys

src/gridfs/file.rs
```diff
@@ -637,11 +637,11 @@ impl GfsFile {
     pub fn to_bson(&self) -> bson::Document {
         let mut doc =
             doc! {
-            "_id" => (self.id.clone()),
-            "chunkSize" => (self.chunk_size),
-            "length" => (self.len),
-            "md5" => (self.md5.to_owned()),
-            "uploadDate" => (self.upload_date.as_ref().unwrap().clone())
+            "_id": self.id.clone(),
```
saghm: and again

src/wire_protocol/operations.rs
```diff
@@ -467,7 +467,7 @@ impl Message {
     /// Returns nothing on success, or an error string on failure.
     pub fn write<W: Write>(&self, buffer: &mut W) -> Result<()> {
         match *self {
-            /// Only the server should send replies
+            // Only the server should send replies
```
saghm: Good catch!

tests/apm/mod.rs
```diff
@@ -35,15 +35,15 @@ fn command_duration() {
 
     let docs = (1..4)
         .map(|i| {
-            doc! { "_id" => i, "x" => (rand::random::<u8>() as u32) }
+            doc! { "_id": i, "x": rand::random::<u8>() as u32 }
```
saghm: Ditto above about multiple keys

tests/apm/mod.rs
```diff
@@ -35,15 +35,15 @@ fn command_duration() {
...
         })
         .collect();
     coll.insert_many(docs, None).unwrap();
     client.add_completion_hook(timed_query).unwrap();
 
     let doc =
         doc! {
-        "$where" => (Bson::JavaScriptCode("function() { sleep(500); }".to_owned()))
+        "$where": Bson::JavaScriptCode("function() { sleep(500); }".to_owned())
```
saghm: and again

tests/client/bulk.rs
```diff
@@ -14,8 +14,8 @@ fn bulk_ordered_insert_only() {
         .map(|i| {
             WriteModel::InsertOne {
                 document: doc! {
-        "_id" => (i),
-        "x" => (i * 11)
+        "_id": i,
```
saghm: God dammit, rustfmt...

I think these ones are bad enough that they should be manually fixed

tests/client/coll.rs
```diff
@@ -693,10 +693,10 @@ fn create_list_drop_indexes() {
     opts1.name = Some("nid".to_owned());
 
     // Test name option
-    let index1 = IndexModel::new(doc!{ "n" => 1, "id" => 1}, Some(opts1));
+    let index1 = IndexModel::new(doc!{ "n": 1, "id": 1}, Some(opts1));
```
saghm: Ditto above about multiple keys

tests/client/coll.rs
```diff
@@ -693,10 +693,10 @@ fn create_list_drop_indexes() {
...
 
     // Test negative value and index name generation
-    let index2 = IndexModel::new(doc!{ "test" => (-1), "height" => 1}, None);
+    let index2 = IndexModel::new(doc!{ "test": -1, "height": 1}, None);
```
saghm: Ditto above about multiple keys

tests/client/coll.rs
```diff
@@ -751,7 +751,7 @@ fn create_list_drop_indexes() {
 
     assert_eq!(2, results.len());
 
-    coll.drop_index(doc!{ "test" => (-1), "height" => 1 }, None)
+    coll.drop_index(doc!{ "test": -1, "height": 1 }, None)
```
saghm: Ditto above about multiple keys

src/cursor.rs
kyeah: yeah, this was a warning that i noticed in the travis build; figured I'd fix it while I'm here, although I wonder why this wasn't noticed or warned about previously

src/gridfs/file.rs
kyeah: SGTM. I'd like to do it in a more selective/incremental manner if possible — I know there have been cases of unintentional slowdowns making that change due to objects with result-wrapper semantics (i.e. they wrap results and basically have two values based on ok/err values), and I can't quite remember if we have anything like that in our codebase.

src/cursor.rs
saghm: From testing it locally, this warning (and the one regarding the doc comment you fixed) only occur on 1.21, not 1.20. It's unclear to me whether this is due to mutability no longer being required with the new compiler/standard library or if the compiler merely improved its existing ability to detect unnecessary mutability qualifiers.

src/auth.rs
kyeah: I tested the latest rustfmt (0.9.0) and it fixed this issue, so I'm going to leave it for now and keep just the relevant macro changes.

tests/client/bulk.rs
kyeah: fixed this one manually.

src/cursor.rs
kyeah: Since the crate builds and the tests pass with 1.20, I'm going to leave this change 👋

src/auth.rs
saghm: Sounds good; as mentioned before "latest rustfmt" is kind of a weird thing right now due to them restarting the versioning a few months back, so I think we should stick to 0.9.0 (the last one before restarting the versioning, which is what I'm using as well).

PR #260, opened 2018-07-02
Handshake spec

saghm:
MongoDB has requested that we implement the [handshake spec](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst) for the driver so that they can identify users of the driver who are connected to Atlas. This is important so that the company can get some actual numbers on who is using the driver and whether it's worth investing more in it. This pull request implements the desired feature.

In addition to Kevin, I'm also asking @pmeredit to take a look at this just to get another pair of MongoDB eyes on the code. (It doesn't seem to allow me to add him as a reviewer because he's apparently not in the `mongodb-labs` organization, hence my @-mentioning him inline).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ env:
     # Use a global cache to re-use dependencies across release builds; downloading a mongodb release is quick.
     - CACHE_NAME=global_cache
   matrix:
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.0.14
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.2.12
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.4.2
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-3.0.15
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-3.2.20
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-3.4.15
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-3.6.5
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-4.0.0
 
 before_install:
     - ./script/start_mongo_release $TRAVIS_OS_NAME $MONGODB_RELEASE
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -2,8 +2,13 @@
 use error::Error::{self, ArgumentError, OperationError};
 use error::Result;
 
+use Client;
+use coll::options::FindOptions;
+use command_type::CommandType;
 use connstring::Host;
+use cursor::Cursor;
 use stream::{Stream, StreamConnector};
+use wire_protocol::flags::OpQueryFlags;
 
 use bufstream::BufStream;
 use std::sync::{Arc, Condvar, Mutex};
@@ -48,6 +53,8 @@ pub struct PooledStream {
     wait_lock: Arc<Condvar>,
     // The pool iteration at the moment of extraction.
     iteration: usize,
+    // Whether the handshake occurred successfully.
+    successful_handshake: bool,
 }
 
 impl PooledStream {
@@ -59,6 +66,11 @@ impl PooledStream {
 
 impl Drop for PooledStream {
     fn drop(&mut self) {
+        // Don't add streams that couldn't successfully handshake to the pool.
+        if !self.successful_handshake {
+            return;
+        }
+
         // Attempt to lock and return the socket to the pool,
         // or give up if the pool lock has been poisoned.
         if let Ok(mut locked) = self.pool.lock() {
@@ -117,7 +129,7 @@ impl ConnectionPool {
     /// Attempts to acquire a connected socket. If none are available and
     /// the pool has not reached its maximum size, a new socket will connect.
     /// Otherwise, the function will block until a socket is returned to the pool.
-    pub fn acquire_stream(&self) -> Result<PooledStream> {
+    pub fn acquire_stream(&self, client: Client) -> Result<PooledStream> {
         let mut locked = try!(self.inner.lock());
         if locked.size == 0 {
             return Err(OperationError(String::from(
@@ -133,6 +145,7 @@ impl ConnectionPool {
                     pool: self.inner.clone(),
                     wait_lock: self.wait_lock.clone(),
                     iteration: locked.iteration,
+                    successful_handshake: true,
                 });
             }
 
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -7,9 +7,11 @@ env:
     # Use a global cache to re-use dependencies across release builds; downloading a mongodb release is quick.
     - CACHE_NAME=global_cache
   matrix:
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.0.14
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.2.12
-    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1204-3.4.2
+    - MONGODB_RELEASE=mongodb-linux-x86_64-ubuntu1404-3.0.15
```
saghm: There are no binaries provided for 12.04 of newer versions of MongoDB, but from a quick inspection, Travis appears to be running 14.04 at this point, so I just updated all of the versions to download the release for that.

src/pool.rs
```diff
@@ -117,7 +122,7 @@ impl ConnectionPool {
     /// Attempts to acquire a connected socket. If none are available and
     /// the pool has not reached its maximum size, a new socket will connect.
     /// Otherwise, the function will block until a socket is returned to the pool.
-    pub fn acquire_stream(&self) -> Result<PooledStream> {
+    pub fn acquire_stream(&self, client: Client) -> Result<PooledStream> {
```
saghm: In order to run `isMaster` with the client metadata after connecting, we need to have a reference to the Client. This unfortunately requires us to pass a reference to it any time `acquire_stream` is called, which percolates through a decent amount of the topology code. Luckily, none of the structs themselves needed to be updated to support this, and the only user-facing API (i.e. `Client#acquire_stream`) can just clone `self` to be passed in, so I think the user-facing changes here are minimal.

src/pool.rs
```diff
@@ -165,4 +172,38 @@ impl ConnectionPool {
...
+            flags,
+            doc! {
+                "isMaster": 1i32,
+                "client": {
+                    "driver": {
+                        "name": ::DRIVER_NAME,
+                        "version": env!("CARGO_PKG_VERSION"),
+                    },
+                    "os": {
+                        "type": ::std::env::consts::OS,
```
saghm: There are a number of additional options fields that could be passed in here, like the version/name of the actual operating system, but given the purpose of this work is just to identify Rust driver usage, I don't think adding the extra complexity beyond something like `Linux (x86_64)` or `Windows (x86_64)` is worth the effort at this point.

tests/client/handshake.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub architecture: String,
+}
+
+#[test]
+fn metadata_sent_in_handshake() {
+    let client = Client::connect("localhost", 27017).unwrap();
+    let db = client.db("admin");
+    let version = db.version().unwrap();
+
+    if version.major < 3 || (version.major == 3 && version.minor < 4) {
```
saghm: The server only recognizes client metadata starting in 3.4. We don't bother checking the version within the driver because the server will silently ignore the metadata in older versions, but the server won't give back the metadata in those versions, so we skip the assertions.

kyeah:
Looks like the repo isn't active on Travis anymore, so we don't have automated tests :(

saghm:
Hmm, I'll check with IT. Good catch!

saghm:
Travis is back online!

saghm:
All passing; ready for review

src/pool.rs
```diff
@@ -141,20 +146,22 @@ impl ConnectionPool {
             if len < locked.size {
                 let socket = try!(self.connect());
                 let _ = locked.len.fetch_add(1, Ordering::SeqCst);
```
kyeah: should this now be incremented after the handshake, in case the handshake fails?

src/pool.rs
```diff
@@ -165,4 +172,38 @@ impl ConnectionPool {
             Err(e) => Err(Error::from(e)),
         }
     }
-}
+
+
+    fn handshake(&self, client: Client, stream: &mut PooledStream) -> Result<()> {
```
kyeah: maybe add a small doc comment here 👋

kyeah: was going to ask if it would make sense to try and DRY out this implementation of `is_master` with `client.is_master` while avoiding exposing the handshake metadata as a public option, but probably more work than it's worth.

src/pool.rs
saghm: Good catch!

saghm:
> Unfortunately this relies on most people using 3.4+ and also upgrading their drivers, but not much we can do there 🤷

Upgrading the driver is definitely a necessity, but the server version isn't a huge deal since this only becomes useful for connections to Atlas, where almost all if not all of the clusters are running on 3.4+.

tests/client/handshake.rs
pmeredit: Are there any other options that check that the version is less than 3.4?  If so, it would make sense to make a function for this.

pmeredit:
Ok, nothing really major here other than it might make sense to at some point add a `VersionAtLeast(version: &[u32]) -> bool` function that compares semantic versions.  I'm not as familiar with how many times you need such functionality in a driver, in the BI-Connector we need it _everywhere_ because of what agg pipeline functions are available in which versions.

tests/client/handshake.rs
saghm: We currently don't have any other tests that only run above a certain version (since most 3.4+ features are not implemented yet), but I like the idea of abstracting this out for future work.

saghm:
> Ok, nothing really major here other than it might make sense to at some point add a version_at_least(version: &[u32]) -> bool function that compares semantic versions.

I've added a macro that skips tests given a minimum version number to meet and used it for the handshake test. If we move to drivers-evergreen-tools/mongo-orchestration, we'd be able to do some macro magic to only define the test based on the environmental variable with the MongoDB version, but for now I think that's overkill.

> Also, does Atlas even support servers older than 3.4?

I don't think so, but it's not a huge deal either way. If there are users using this driver on Atlas, I'm not super worried that they'd be hesitant to upgrade past 3.2 even if it's not required.

src/pool.rs
```diff
@@ -140,21 +145,23 @@ impl ConnectionPool {
...
+                let mut stream = PooledStream {
                     socket: Some(socket),
                     pool: self.inner.clone(),
                     wait_lock: self.wait_lock.clone(),
                     iteration: locked.iteration,
-                });
+                };
+
+                self.handshake(client, &mut stream)?;
+                let _ = locked.len.fetch_add(1, Ordering::SeqCst);
```
kyeah: I thought about this more and realized an interesting issue here — the PooledStream is returned to the pool via the `Drop` trait, which means that if the handshake fails, 
- `acquire_stream` returns an error instead of the stream
- the stream is dropped
- the new socket is added into the pool of available connections anyways

If we're okay with the possibility of a socket being opened _without_ a successful handshake, we should increment this _after_ the PooledStream is created, but before the handshake.

If we're not okay with that possibility, we need to do extra work — either add a "completed_handshake" kinda boolean flag to pooled streams, or figure out something clever where we don't have to create the PooledStream until we handshake.

saghm: Another good catch! I went with the boolean flag on the stream; I don't think the overhead of a few extra booleans should make a noticeable difference, and I think it's important to make sure that every connection handshakes as the spec requires. (I also didn't rebase this time so that the changes are easier to follow; I'll just rebase to clean things up before merging once we're done making changes)

saghm:
Waiting for a final Travis run after rebasing, and then I'll merge!

On a tangential note, it looks like 0.3.8 was never published to crates.io. Since it was already tagged and we don't want to mess with history, we shouldn't add this to 0.3.8. Should I just release the version with this as 0.3.9 (and if so, should we still publish 0.3.8)?

kyeah:
yep, let's release both versions 👍

saghm:
Hmm, actually, it looks like we didn't actually tag 0.3.8, so I'll just include this in that release.

PR #265, opened 2018-07-13
WIP: add batch_size option to command_cursor

nevi-me:
I don't know if I've done it correctly, there are 2 aggregation tests that are failing, I'll look into them

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -104,7 +104,7 @@ impl Collection {
         };
 
         let mut read_preference = self.read_preference.clone();
-        let mut batch_size = 1;
+        let mut batch_size: Option<i32> = None;
 
         match options {
             Some(aggregate_options) => {
@@ -112,7 +112,7 @@ impl Collection {
                     read_preference = read_preference_option.clone();
                 }
 
-                batch_size = aggregate_options.batch_size;
+                batch_size = Some(aggregate_options.batch_size);
 
                 spec = merge_options(spec, aggregate_options);
             }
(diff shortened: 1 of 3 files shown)


src/coll/mod.rs
```diff
@@ -104,13 +104,16 @@ impl Collection {
         };
 
         let mut read_preference = self.read_preference.clone();
+        let mut batch_size = 1;
```
saghm: I'd just make this an option initially set to `None` and then changed to `Some(batch_size)` on line 115.

src/coll/mod.rs
```diff
@@ -1029,6 +1033,7 @@ impl Collection {
             cmd,
             CommandType::ListIndexes,
             self.read_preference.to_owned(),
+            Some(1)
```
saghm: Just pass `None` here; the idea behind having the function in the cursor module accept an `Option` as a parameter is to avoid having to specify the default elsewhere.

src/cursor.rs
```diff
@@ -96,6 +96,7 @@ impl Cursor {
     /// `doc` - Specifies the command that is being run.
     /// `cmd_type` - The type of command, which will be used for monitoring events.
     /// `read_pref` - The read preference for the query.
+    /// `batch_size` - The number of records to return per batch
```
saghm: I'd add a note that specifies that the default is 1 to facilitate sending commands

saghm:
Hmm, from taking a look at the failed test, I'm actually thinking the correct solution is to have the aggregate method call `Database::command` rather than `Database::command_cursor` (and then reverting the other changes). If you're up for it, you can make the changes to see if that works; otherwise, I can do it.

nevi-me:
I'm struggling to find a way of changing from `command_cursor` to `cursor`. I'll leave it to you if you don't mind.

PR #266, opened 2018-07-18
Replaced rust-crypto with RustCrypto crates

newpavlov:
`rust-crypto` is no longer maintained, plus monolithic architecture means that you pull a lot of stuff which will not be used. [RustCrypto](https://github.com/RustCrypto) crates provide a significantly more lightweight (and protable) alternative without sacrificing performance, as most of the code is based on `rust-crypto` implementation with additional optimizations.


kyeah:
Perfect! thank you for taking the time to make this update 🤗

Issue #245, opened 2017-12-04
Aggregate batch size reset after first batch


polachok:
I'm using this code:

```
		let mut options = AggregateOptions::new();
		options.batch_size = 100;
		let mut cur = coll.aggregate(req, Some(options))?;
		while cur.has_next()? {
			let batch = cur.drain_current_batch()?;
			println!("BATCH: {:?}", batch.len());
			for item in batch {
                                 ...
                        }
                 }
```

I have around ~3000 records in this collection
What I expect: batch len would be 100 on every iteration.
What I see: batch len is 100 first time, and then 1.
This makes it very slow (several minutes  for 3000 records), if I set batch to 5000, I get them all in under 5 seconds.

nevi-me:
@saghm #236 addresses the problem, but looks like it still exists on 0.3.9

Is there a workaround? 
I have something like the below

```rust
    let mut aggregation_options = mongodb::coll::options::AggregateOptions::new();
    aggregation_options.allow_disk_use = Some(true);
    aggregation_options.batch_size = 100;
    aggregation_options.use_cursor = Some(true);
```

saghm:
Thanks for reporting this @polachok! That's the correct workaround @nevi-me (although I don't think you need to use `use_cursor`, as aggregations use cursors by default except on old versions of the database). 

I'll look into the issue to see why this is happening; offhand, it seems like it's probably that the batch size is not propagated to the `getMore` commands (which are what drivers use to get batches after the first).

nevi-me:
Thanks @saghm, I noticed that even if I don't `use_cursor` it still uses it by default, so that makes sense. Could this be it? https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/01af46789f95d1101e93ca7ea70200b6f4a62eb4/src/cursor.rs#L111

saghm:
Hmm, yes, that looks like it's the issue. Normally when sending a command, there aren't any batches in the result, but given that queries are still send as commands, that needs to be overridden in some cases. I think the best fix would be to have that method take a `batch_size: Option<i32>` parameter and then having that assignment be `options.batch_size = batch_size.unwrap_or(1)`. I can make this change tonight if you're willing to wait, but feel free to make a PR if you'd like the change sooner!

nevi-me:
[...]

PR #268, opened 2018-07-20
fix getMore batch size

saghm:
This fixes https://github.com/mongodb-labs/mongo-rust-driver-prototype/issues/245 (for real this time!)

After playing around with a few more complicated fixes, I realized that the best solution was dead simple: the batch size of the cursor should just be the the size of the first batch returned by the server. While this technically means that the cursor will contain an "incorrect" batch size when the result set is smaller than the calculated batch size, this won't cause any issues in practice since the field is private and by definition no more batches will be fetched.

For testing, the `Cursor::drain_current_batch` method conveniently lets us get the actual batch size without needing to expose any additional details. I verified that the tests failed without the fix being added, so I think we fixed the issue for real this time.

cc @nevi-me

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -405,7 +405,7 @@ impl Cursor {
         Ok(Cursor {
             client: client,
             namespace: namespace,
-            batch_size: options.batch_size.unwrap_or(DEFAULT_BATCH_SIZE),
+            batch_size: buf.len() as i32,
             cursor_id: cursor_id,
             limit: options.limit.unwrap_or(0) as i32,
             count: 0,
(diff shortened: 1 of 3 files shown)


nevi-me:
Thanks @saghm, I was planning on tinkering around over the weekend.

In terms of SEMVER, would this be a patch?

saghm:
Since it's still 0.x, we can technically make it anything we want! I'm not sure whether I think this should be 0.4.0 or 0.3.10; it is just a trivial bugfix in terms of the changed code, but given the severity of the bug, it is a pretty significant change. I'd be interesting in hearing @kyeah's opinion on how we should bump the version.

nevi-me:
Makes sense, I'd prefer 0.3.10.

I've been porting some interesting aggregation thing from Kotlin to Rust, I'm excited that I'm a few steps closer to getting it working.

Issue #278, opened 2018-09-26
Read Concerns and Write majority


stefano-pogliani:
Hello,

I was looking at support for `majority` [read](https://docs.mongodb.com/manual/reference/read-concern-majority/) and [write](https://docs.mongodb.com/manual/reference/write-concern/#writeconcern.%22majority%22) concern.

It seems the driver has support for [read modes](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/master/src/common.rs#L11) which I could not find in the [MongoDB documentation](https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/) but not read *concerns* (which were introduced in MongoDB 3.2).

As for write concerns the driver has support for a numeric `w` parameter but does not seem to have support for the `"majority"` or `<tag set>` concerns.

Would there be interest in a PR to support read concerns and extend write concerns?

I would be happy to work on it but it would be likely to include breaking changes (I'm not sure about the read concerns yet but the `w` parameter would need to change from `i32` to - probably - an `enum`).
Would such breaking change be a deal breaker?

saghm:
Hello everyone! We've just released an alpha of [an official MongoDB Rust driver](https://github.com/mongodb/mongo-rust-driver); you can read about it [here](https://blog.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver)! Because of this, we won't be updating this driver anymore, and the repository will be made read-only as soon as I get finished posting these messages. Thanks for using this driver, and we hope you try out the new one!

saghm:
Apologies, the blog post link is this: https://www.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver

PR #280, opened 2018-09-29
Implement support for read concerns

stefano-pogliani:
This PR addresses the read concern support part of #278.

**Note**: this PR introduces a breaking change because it requires the default read concern and max time ms attributes to be passed to some methods.

### DIFF
--- a/script/start_mongo_release
+++ b/script/start_mongo_release
@@ -16,9 +16,17 @@ if [ ! -d "$install_dir/$release" ]; then
 fi
 
 # Start mongod processes
+enable_read_concern=""
+if [[ "$release" == *"3.2."* ]]; then
+  enable_read_concern="--enableMajorityReadConcern"
+fi
+if [[ "$release" == *"3.4."* ]]; then
+  enable_read_concern="--enableMajorityReadConcern"
+fi
+
 data="$install_dir/data/$release"
 data_ssl="$data-ssl"
 
 mkdir -p $data $data_ssl
-$install_dir/$release/bin/mongod --fork --dbpath $data --syslog --port 27017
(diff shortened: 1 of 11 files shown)


src/coll/mod.rs
```diff
@@ -43,10 +45,13 @@ impl Collection {
         db: Database,
         name: &str,
         create: bool,
+        max_time_ms: Option<i64>,
```
saghm: In retrospect,everything beyond the `Database` and Name should probably be abstracted to an options struct (and ditto for creating a database only needing `Client` and name). Are you interested in taking a stab at this, or would you prefer that I make the change first and then you can rebase it?

src/common.rs
```diff
@@ -6,6 +6,62 @@ use bson::{self, Bson};
 use std::collections::BTreeMap;
 use std::str::FromStr;
 
+/// Level of desired consistency and isolation properties of the data read.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
saghm: Out of curiosity, what's the rationale for deriving ordering on `ReadConcern`? I don't really think of read concerns has having a natural order, and I'm not sure that anyone would be able to intuitively guess the ordering of a given pair of read concerns.

script/start_mongo_release
```diff
@@ -16,9 +16,17 @@ if [ ! -d "$install_dir/$release" ]; then
 fi
 
 # Start mongod processes
+enable_read_concern=""
+if [[ "$release" == *"3.2."* ]]; then
+  enable_read_concern="--enableMajorityReadConcern"
```
saghm: Is there a reason we need the tests to run with majority read concern?

saghm (commented):
Thanks for the PR, and apologies for taking so long to review this! If you're still interested in working on this, I have a couple of questions and suggestions. If not, I totally understand, and I can make the additional changes and get it merged.

stefano-pogliani:
Hi @saghm and thank you for the feedback.
I'll be happy to look at your comments but I don't think I'll have the time this week.

saghm:

PR #283, opened 2018-10-19
fix the implementation of the read function for gridfs/File

magiclen:
This pull request solved a problem for the implementation of the `read` function of the `File` struct in the `gridfs` module.

When the **OperationError(String::from("Chunk not found.")** cache state is set, and the output buffer(buf) is bigger than the read buffer(rbuf), the `get_chunk` function should be called only when there are more than one remaining chunks. Or else, if there isn't any remaining chunk, the `read` and other `read*` functions will return **OperationError(String::from("Chunk not found.")**.

### DIFF
--- a/src/gridfs/file.rs
+++ b/src/gridfs/file.rs
@@ -547,7 +547,7 @@ impl io::Read for File {
         }
 
         // Read all required chunks into memory
-        while self.rbuf.len() < buf.len() && (self.chunk_num as i64) * (self.doc.chunk_size as i64) < self.doc.len{
+        while self.rbuf.len() < buf.len() && (self.chunk_num as i64) * (self.doc.chunk_size as i64) < self.doc.len {
             let chunk = self.get_chunk()?;
             self.rbuf.extend(chunk);
         }
--- a/src/gridfs/file.rs
+++ b/src/gridfs/file.rs
@@ -547,7 +547,7 @@ impl io::Read for File {
         }
 
         // Read all required chunks into memory
-        while self.rbuf.len() < buf.len() {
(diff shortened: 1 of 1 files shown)


saghm:
To be honest, I'm honestly not sure the pre-loading of the next chunk is really worth it. Most of the official drivers don't do this as far as I'm aware, and I don't think the [gridfs spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst) itself says anything about it. We also have had issues with it in the past (like the one fixed in https://github.com/mongodb-labs/mongo-rust-driver-prototype/pull/158). This is obviously out of scope for this PR though, so there's no need to implement it before merging this.

magiclen:
I've added the stupid whitespace and also took a look at the preloading logic. I didn't see any possible race condition.

kyeah:
> To be honest, I'm honestly not sure the pre-loading of the next chunk is really worth it. Most of the official drivers don't do this as far as I'm aware, and I don't think the gridfs spec itself says anything about it. We also have had issues with it in the past (like the one fixed in #158). This is obviously out of scope for this PR though, so there's no need to implement it before merging this.

Oh interesting...my recollection was that the gridfs spec noted/recommended to do this — or the C driver did this — but if not we can probably remove it.

kyeah:
Thanks for the contribution @magiclen.

saghm:
Interestingly, the C driver actually explicitly doesn't implement the GridFS spec due to a pre-existing API that it can't make breaking changes to. I think sticking to the spec is probably preferable to following the C driver on this one.

PR #290, opened 2018-11-12
Re-export bson lib

### LINKED ISSUE #281: Please re-export bson lib.
It would be really awesome if this create would `pub use bson;` from the root of the library. Here is some reasoning as to why:
1. every time this library releases a new version (even patch versions) where the `bson` dependency changes, **everything breaks.** Users of this library must (currently) depend on the BSON crate (because you need to in order to use the mongo crate).
    - the best you can do to mitigate this issue is to require the closest `minor` version of `bson` which matches this crate's dep. When the bson dep is updated from, say, `0.10` -> `0.12`, this doesn't help.
    - having the dep as `bson = "*"` does not help either, because if bson `0.13` is released, and this create still requires `0.12`, then things still break.
2. BSON is not used that often outside of mongodb context. So ... why not just re-export it from this lib? Users could still directly depend on it, of course, even if it is re-exported from this crate.

This issue presents as follows:
```
error[E0308]: mismatched types
   --> myfile.rs:666:666
    |
666 |         ... blah blah some code ...
    |
    = note: expected type `bson::ordered::OrderedDocument` (struct `bson::ordered::OrderedDocument`)
               found type `bson::ordered::OrderedDocument` (struct `bson::ordered::OrderedDocument`)
note: Perhaps two different versions of crate `bson` are being used?
```
Error message is spot on, and very helpful. But there is still a compilation error.

----

Once re-exported by this crate, users would bring it into scope via `use mongodb::bson;` and would not even need to declare it as an additional dependency. 

As always, thanks for the great work on this lib. I'm stoked to see it continue to evolve. Hoping to help out a bit more in the future myself. Especially to get some async i/o support in place.

kyeah:
Closes #281.

This converts usage of `macro_use` into explicit imports of `use bson::{bson, doc}`.

Also re-exports bson::* to remove downstream requirements of installing the correct version of bson.

Cut as version 0.3.11.

cc. @H2CO3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ license = "Apache-2.0"
 name = "mongodb"
 readme = "README.md"
 repository = "https://github.com/mongodb-labs/mongo-rust-driver-prototype"
-version = "0.3.10"
+version = "0.3.11"
 
 [dependencies]
 bitflags = "1.0.0"
--- a/README.md
+++ b/README.md
@@ -22,8 +22,7 @@ The driver is available on crates.io. To use the MongoDB driver in your code, ad
 
 ```toml
 [dependencies]
-bson = "0.12.2"
-mongodb = "0.3.10"
+mongodb = "0.3.11"
 ```
 
 Alternately, you can use the MongoDB driver with SSL support. To do this, you must have OpenSSL installed on your system. Then, enable the `ssl` feature for MongoDB in your Cargo.toml:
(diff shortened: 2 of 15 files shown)


H2CO3:
Awesome! 👏

Kampfkarren:
This is vital to me, any ETA on merge?

H2CO3:
Awesome, you guys are the best! 👍

PR #292, opened 2018-11-15
Allow SSL configurations to fall back on the system CA

saghm:
The current SSL API requires users to specify a certificate authority file. All of the official drivers allow users to instead fallback to the system CA, so we should do that too.

In order to test this properly, we'd need to spin up a third mongod that requires SSL but does not have a CA specified and then make a specific test that attempts to connect without specifying it. Given that this is a very small variation on the existing SSL tests and that we rely entirely on the OpenSSL implementation to perform this feature, I don't think it's worth increasing our Travis usage by 50% for this (especially given that Travis servers are grouped by Github organization/user, and there are a _lot_ of projects in mongodb-labs, some of which have more pressing needs for prompt CI feedback). I've verified that this works locally.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -257,7 +257,7 @@ impl ClientOptions {
     #[cfg(feature = "ssl")]
     /// Creates a new options struct with a specified SSL certificate and key files.
     pub fn with_ssl(
-        ca_file: &str,
+        ca_file: Option<&str>,
         certificate_file: &str,
         key_file: &str,
         verify_peer: bool,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -257,7 +257,7 @@ impl ClientOptions {
     #[cfg(feature = "ssl")]
     /// Creates a new options struct with a specified SSL certificate and key files.
     pub fn with_ssl(
-        ca_file: &str,
+        ca_file: Option<&str>,
```
saghm: I realize that this is a breaking API change, but given the pre-1.0 version, it's technically semver compliant. While we could get around this by defining another pair of methods where the CA file isn't required, I think this one-time API change is worth not having four separate constructors for a relatively simple struct, given that the fix will be very trivial for users.

goriunov:
@saghm  When should we expect this fix to merge. I am currently testing library and having issues to connect to MongoDB Atlas. 

Thank you for your work.

saghm:
@goriunov I just merged it, and pending a final test run, I'll release it right away!

saghm:
Published in 0.3.12

PR #299, opened 2019-01-07
A few bug fixes & two convenience items.

### LINKED ISSUE #224: Should have a method for listing index models.
Currently there is a method for [listing indices on a collection](https://docs.rs/mongodb/0.3.0/mongodb/coll/struct.Collection.html#method.list_indexes), which returns a `Result<Cursor<bson::ordered::OrderedDocument>>`. This is great, however it would be awesome to have a method which returns `Result<Cursor<IndexModel>>` for interoperability with the other index methods dealing with `IndexModel`.

I'm happy to open a PR for this as well.

### LINKED ISSUE #282: IndexOptions.storage_engine appears to have an incorrect type.
Hey guys, I hope all is well.

I've been in the process of building [wither](https://github.com/thedodd/wither) version `0.6.0`, implementing a custom derive for creating models. [PR is here](https://github.com/thedodd/wither/pull/16) for reference.

During the process, I've written extensive tests for the index system. I've discovered that the [IndexOptions.storage_engine](https://docs.rs/mongodb/0.3.10/mongodb/coll/options/struct.IndexOptions.html#structfield.storage_engine) field is of type `Option<String>`, but mongo appears to be expecting a BSON document & [the docs seem to corroborate this](https://docs.mongodb.com/manual/reference/method/db.collection.createIndex/#ensureindex-options). To be sure, I also tried embedding a JSON document inside of a string and passing that as the value, but I ended up getting the same error. Here is the error I'm seeing:

```rust
DefaultError("Failed to create index: \"storageEngine\" options must be a document if present")
```

As the data type determines what is sent across the wire, there doesn't seem to be a workaround. We'll have to update the data type.

Please let me know if I am missing or overlooking anything, and thanks for all the work on this crate! Cheers!

### LINKED ISSUE #289: Driver is not allowing creation of a geoHaystack index.
Looks like the code here is not allowing for it: https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/master/src/coll/options.rs#L366-L377

GeoHaystack indices are described in the official MongoDB docs here: https://docs.mongodb.com/manual/core/geohaystack/

Should be a pretty simple fix.

thedodd:
Closes #224
Closes #282
Closes #289

Fixes a bug where `geoHaystack` indexes were being blocked from creation.

Fixes a bug where the `storageEngine` index option had the wrong type.

Implemented a convenience method as discussed in #224. `Cursor` is still
the base iterator type, but the output has been wrapped as an `impl
Iterator` for convenience. Works exactly the same way, but yields
`IndexModel` elements instead of plain old `Document`s.

Added serde derivations to `IndexModel` & (by necessity) `IndexOptions`.
The BSON crate already directly depends on the serde, serde_derive &
serde_json. They've been added to the dependencies section of the
Cargo.toml as they are now direct dependencies of this crate, but this
does not technically add any new deps.

----

### thoughts on semver version to use
- nothing backwards incompatible here. #282 may seem like a backwards incompatibility, but as the server was rejecting any previously given values, I would still consider this to be a patch increment change.
- we could totally do a `0.4` release. No reason not to really. This changeset does add `list_index_models` & serde derivations on `IndexModel`.

Any reason not to do a `0.4`?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ data-encoding = "2.0.0"
 rand = "0.3.15"
 scan_fmt = "0.1.1"
 semver = "0.8.0"
+serde = "1"
+serde_derive = "1"
 separator = "0.3.1"
 time = "0.1.37"
 md-5 = "0.7.0"
@@ -26,7 +28,6 @@ sha-1 = "0.7.0"
 hmac = "0.6.2"
 pbkdf2 = "0.2.0"
 hex = "0.3.2"
-
 
 [dependencies.clippy]
 optional = true
@@ -36,18 +37,16 @@ version = "~0"
 optional = true
 version = "0.10.15"
 
+[dependencies.serde_json]
+version = "1"
+features = ["preserve_order"]
+
 [dependencies.textnonce]
 default-features = false
 version = "0.6.3"
 
 [dev-dependencies]
 approx = "0.1.1"
-serde = "1.0.69"
-serde_derive = "1.0.69"
-
-[dev-dependencies.serde_json]
-features = ["preserve_order"]
-version = "1.0.3"
 
 [features]
 default = []
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -1054,12 +1054,7 @@ impl Collection {
     /// This is the same as `list_indexes`, and still uses a `Cursor` under the hood. The elements
     /// are serialized as `IndexModel`s as they are received.
     pub fn list_index_models(&self) -> Result<impl Iterator<Item=Result<IndexModel>>> {
-        let cmd = doc!{ "listIndexes": self.name() };
-        self.db.command_cursor(
-            cmd,
-            CommandType::ListIndexes,
-            self.read_preference.to_owned(),
-        ).map(|cursor| {
+        self.list_indexes().map(|cursor| {
             cursor.map(|doc_res| {
                 doc_res.and_then(|doc| -> Result<IndexModel> {
                     bson::from_bson(bson::Bson::Document(doc)).map_err(|err| DecoderError(err))
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -18,7 +18,7 @@ use cursor::Cursor;
 use db::{Database, ThreadedDatabase};
 
 use Result;
-use Error::{ArgumentError, ResponseError, OperationError, BulkWriteError};
+use Error::{ArgumentError, DecoderError, ResponseError, OperationError, BulkWriteError};
 
 use wire_protocol::flags::OpQueryFlags;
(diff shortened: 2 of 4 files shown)


thedodd:
@saghm per our discussion in #224, I've implemented the new `list_index_models` method on `Collection`. Using a return type of `Result<impl Iterator<Item=Result<IndexModel>>>`. Still uses the underlying `Cursor` as the base iterator, of course, but then just maps the values as they are received and such. 

@kyeah this resolves the bug we discussed in #282.

Also resolves the subtle issue outlined in #289.

thedodd:
@saghm && @kyeah hey guys, I know you are super busy, and I respect that, I am in the same situation often times.

I was wondering how you guys would feel about extending ownership/admin permissions to me on this crate. I am more than happy to help maintain it and keep things moving forward. There are a bunch of really big innovations coming down the Rust pipeline with async/await, and a bunch of features that many folks would love to see implemented in this crate (async, transactions, changestreams &c).

Anyway, let me know what you think. I am definitely not trying to push or pry, just looking to help out. I use this crate pretty heavily.

thedodd:
@saghm && @kyeah just wanted to ping you guys again since its been about a month. I am happy to jump on a google hangout with you guys and discuss these changes, as well as long term plans for this project and so forth.

Please let me know. I want to start work on a tokio based version of this crate, but I don't want to fork ... as that would split the community around this crate too much, and I want to knock out a few of the other issues here first.

Cargo.toml
```diff
@@ -19,6 +19,8 @@ data-encoding = "2.0.0"
 rand = "0.3.15"
 scan_fmt = "0.1.1"
 semver = "0.8.0"
+serde = "1"
+serde_derive = "1"
```
thedodd: Just FYI, these two dependencies were already being compiled as standard dependencies because they are required by other deps. So these are not actually new dependencies, they are now just "top-level" dependencies.

saghm: Yep, `bson` already pulls those in. Thanks for the heads up!

thedodd:
Hey @kyeah just a friendly ping to see if you might have time to take a look here.

thedodd:
@kyeah & @saghm you guys want me to update the changelog before merging this? I'm happy to do so.

kyeah:
@thedodd If you can add a changelog and bump the version in a new PR that would be great — I'll merge this so the version bump lives on its own commit.

thedodd:
@kyeah negative, I do not have merge permissions.

kyeah:
@thedodd merged into master; add a changelog/bump the version to 0.4 and we can run a release. Thank you!

thedodd:
Woot woot! Will do, boss. Should be able to knock that out tomorrow morning.

Issue #213, opened 2017-05-13
How to set pool size for my client?


YjyJeff:
Hi, I find that the default pool size is 5. I think it's too small. How can I set the pool size ? 
Thanks in advance!

saghm:
Thanks for filing the issue! It actually looks like we don't currently have a publicly accessible way of doing this right now, so we'll need to fix that!

 @kyeah, should we add an option to `ClientOptions` and pass that through all the way through to `ConnectionPool::with_size`?

kyeah:
yep, that would be the way to go here! feel free to assign yourself if you're thinking of picking up this issue 🙏

saghm:
Will do! I have a train ride tomorrow that's a few hours long, so I might take a look at it then

saghm:
Hi! I just wanted to give an update on this since it's been a while without any visible progress on this. Upon looking into the client pooling in the driver, I found that there were some mistakes in the design that caused more clients than expected to be created (with the excess ones not being accessible to the user and therefore being dead weight), and we've decided to try to rewrite the connection pooling to use [r2d2](https://crates.io/crates/r2d2), which has built in support for setting a custom pool size. The first step in this work is #218, although we'll need at least one more PR to complete this work.

Let me know if you have any questions about all this!

saghm:
The pool size determines the number of connections that are able to be open to a given server at once.

saghm:
Hello everyone! We've just released an alpha of [an official MongoDB Rust driver](https://github.com/mongodb/mongo-rust-driver); you can read about it [here](https://blog.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver)! Because of this, we won't be updating this driver anymore, and the repository will be made read-only as soon as I get finished posting these messages. Thanks for using this driver, and we hope you try out the new one!

saghm:
Apologies, the blog post link is this: https://www.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver

PR #304, opened 2019-03-13
Added Connection Pool Size and Idle Connection Timeout to ClientOptions

### LINKED ISSUE #247: Reference cycle of `Client` and `Monitor` leaks connection
There seems to be a connection leak when creating a `Client`:
```rust
extern crate mongodb;
use mongodb::Client;
use mongodb::ThreadedClient;

fn main() {
  let client = Client::connect(host, port).unwrap();
  // This will print 2 instead of 1.
  println!("{}", std::sync::Arc::strong_count(&client));
}
```

I believe this is caused by the `Monitor` instance stored in the `Server` instances that are added to the `Topology` stored in the `Client`:

  * [Once the client is created the servers are added](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/master/src/lib.rs#L387)
  * [Each server keeps a reference to a monitor](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/0f348beb955a9d645d25c854ff5aff149befc526/src/topology/server.rs#L89)
  * [The monitor holds the client for when it refreshes the topology](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/0f348beb955a9d645d25c854ff5aff149befc526/src/topology/monitor.rs#L71)

I would be happy to open a Pull Request to change the `Client` to a `std::sync::Weak<ClientInner>`.
With this change the cycle is broken but it means adding two `self.client.upgrade().unwrap()` to:

  * https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/0f348beb955a9d645d25c854ff5aff149befc526/src/topology/monitor.rs#L272
  * https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/0f348beb955a9d645d25c854ff5aff149befc526/src/topology/monitor.rs#L329

Would this approach be acceptable?

richerarc:
I added several options / fixed some bug that were blocking me and my team to use this driver in production.

This fixes #247 , #213 and an other issue were connection held in the ConnectionPool were never de-allocated unless the client was dropped.

I've never contributed to this project before so feel free to ask for improvement / code change, I will do as requested to fit within the code style of the repo


Igosuki:
noice

phantomxe:
notice

Liby99:
I would really want this to be merged...

saghm:

PR #305, opened 2019-03-19
Set TCP_NODELAY for the socket

### LINKED ISSUE #277: Performance is worse than terrible
mongodb = "0.3.10"

rustc --version
rustc 1.30.0-nightly (4591a245c 2018-09-22)

mongod --version
db version v4.0.2

Rust:
https://gist.github.com/inv2004/87b49ee4217872d9260e13a09e6bf524
```
$ time ./target/release/bench_mongo
real    0m44.043s
```

Go:
https://gist.github.com/inv2004/d8e397ee7d0553940b60dff7f3005540
```
$ time ./bench_mongo
real    0m0.798s
```

---added---
mongo_driver:
real 0m0.583s

athre0z:
This greatly increases throughput on many platforms and resolves #277

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -91,7 +91,11 @@ impl StreamConnector {
 
     pub fn connect(&self, hostname: &str, port: u16) -> Result<Stream> {
         match *self {
-            StreamConnector::Tcp => TcpStream::connect((hostname, port)).map(Stream::Tcp),
+            StreamConnector::Tcp => {
+                let stream = TcpStream::connect((hostname, port))?;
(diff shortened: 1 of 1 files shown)


athre0z:
I additionally implemented buffering for the read half in order to further improve performance (by about 2x for my personal use case and workload). I'm not sure how this is best done for the SSL version since SslStream doesn't support splitting / cloning. Buffering the write half might be a good idea as well, however this is a lot more error prone to implement since we'd have to ensure correct flushing in all situations, so I didn't do it for now.

PR #316, opened 2019-06-19
Added dyn keyword for trait objects

jblondin:
'dyn <trait>' syntax introduced in Rust 1.27 for trait objects. Updated to use recommended syntax.

### DIFF
--- a/src/wire_protocol/header.rs
+++ b/src/wire_protocol/header.rs
@@ -115,7 +115,7 @@ impl Header {
     /// # Return value
     ///
     /// Returns nothing on success, or an Error on failure.
-    pub fn write(&self, buffer: &mut Write) -> Result<()> {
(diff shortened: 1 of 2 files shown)


src/wire_protocol/header.rs
```diff
@@ -115,7 +115,7 @@ impl Header {
     /// # Return value
     ///
     /// Returns nothing on success, or an Error on failure.
-    pub fn write(&self, buffer: &mut Write) -> Result<()> {
+    pub fn write(&self, buffer: &mut dyn Write) -> Result<()> {
```
saghm: We should probably change these to be generic rather than use trait objects, but that's not necessary for this PR.

jblondin: Good point, updated

PR #319, opened 2019-07-19
adding arrayFilters feature

phantomxe:


saghm:

PR #324, opened 2019-08-23
Allow openssl-sys vendoring for static musl builds

chastabor:
This change allows the user to take advantage of the vendor feature included in the openssl-sys crate that openssl depends on. The vendor feature helps when using musl to statically build the mongodb driver with openssl. When users include the "vendor" feature while using the mongodb "ssl" feature, they may use the rust build container from docker hub without also needing to compile the openssl library before starting their cargo build/install process.

An example Dockerfile would be as follows:

> FROM rust:1.37.0
> WORKDIR /usr/src
> 
> _#_ Download the target for static linking.
> RUN apt-get update \\
>   && DEBIAN_FRONTEND=noninteractive apt-get -y install ca-certificates musl-dev musl-tools make automake \\
>   && update-ca-certificates \\
>   && rustup target add x86_64-unknown-linux-musl \\
>   && mkdir /usr/src/\<project>
> 
> WORKDIR /usr/src/\<project>
> COPY Cargo.* ./
> COPY src ./src
> RUN cargo install --target x86_64-unknown-linux-musl --path .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,10 @@ version = "~0"
 optional = true
 version = "0.10.15"
 
+[dependencies.openssl-sys]
+optional = true
+version = "*"
+
 [dependencies.serde_json]
 version = "1"
 features = ["preserve_order"]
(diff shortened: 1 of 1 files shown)


saghm (approved):
This looks good to me! @kyeah?

saghm:

PR #326, opened 2019-09-24
Implement byte length calculation that doesn't serialize the bson object

jcdyer:
Speed up the byte_length calculation by not serializing the bson data structure.  Correct sizes were taken from http://bsonspec.org/spec.html, and verified by running the test suite against both implementations simultaneously (see first commit).

I ran benchmarks that test the old and new `ByteLength for bson::Document`  implementations with:

1. An empty Document (empty_doc)
2. A Document containing 10,000 string records with keys `"key 0"` through `"key 9999"` and values `Bson::String("lorem ipsum dolor sit amet")` (broad_doc)
3. A nested document 10,000 documents deep (`doc!{"key 0": doc!{"key 1": doc!{"key 2": ...}}}`) (deep_doc).

document type  | old (serializing) | new (calculating) | change 
------------- | ---------------- | ----------------- | ------
empty_doc | 61.6 ns | 9.8ns | -84.1%
broad_doc | 391.7 us | 105.1us | -73.2%
deep_doc | 41.1 ms | 978.3 us | -97.6%

The benchmarks are not included in the PR, because ByteLength is not a publicly accessible trait, so I wrapped them in a temporary public function that the benchmark calls.  Benchmark source is provided below:

## `src/wire_protocol/operations.rs`

```rust
pub fn bl(doc: &bson::Document) -> i32 {
    doc.byte_length().unwrap()
}
```

## `benches/byte_length_test.rs`

```rust
#[macro_use]
extern crate criterion;

#[macro_use]
extern crate bson;

extern crate mongodb;

use criterion::{Criterion};

use mongodb::wire_protocol::operations::bl;

mod docs {
    pub(super) fn empty_doc() -> bson::Document {
        bson::Document::new()
    }

    pub(super) fn big_doc() -> bson::Document {
        let mut doc = bson::Document::new();
        for n in 0..10000 {
            doc.insert(format!("key {}", n), "lorem ipsum dolor sit amet");
        }
        doc
    }

    pub(super) fn deep_doc() -> bson::Document {
        let mut doc = bson::Document::new();
        for n in 0..10000 {
            let mut newdoc = bson::Document::new();
            newdoc.insert(format!("key {}", n), doc);
            doc = newdoc;
        }
        doc
    }
}

fn byte_length_benchmark(c: &mut Criterion) {
    let d = docs::empty_doc();
    c.bench_function("empty_doc_len", |b| b.iter(|| bl(&d)));
}

fn byte_length_big(c: &mut Criterion) {
    let d = docs::big_doc();
    c.bench_function("big_doc_len", |b| b.iter(|| bl(&d)));
}

fn byte_length_deep(c: &mut Criterion) {
    let d = docs::deep_doc();
    c.bench_function("deep_doc_len", |b| b.iter(|| bl(&d)));
}

criterion_group!(
    benches,
    byte_length_empty,
    byte_length_big,
    byte_length_deep,
);
criterion_main!(benches);

```


saghm:

Issue #301, opened 2019-01-17
raw bson


univerz:
sometimes it's not possible (https://jira.mongodb.org/browse/SERVER-267) or practical (database load) to project only relevant fields (or if you want to just copy objects between servers). deserialization of bson is then the performance bottleneck.

it would be great if `Cursor` could yield `raw_bson` with behaviour like in c++ driver.

for my data, it's <10seconds vs 3minutes.

univerz:
> I'm assuming it is just the raw bytes coming off the wire. That sounds about right?

yes, sorry for the confusion

> Would you want to have this interface on the collection level, or would you want this to exist on the database level as well (for db commands and such)? We should take a look at the code and determine what the level of effort would be so that we can put an implementation plan together.

for my current use case, collection level is sufficient. as i'm looking through the source code, this looks like a serious task as bson crate is used everywhere. please consider this only as an input for the future work if/when mongodb (company) decides to officially support this driver.

bson crate is nice and convenient, but maybe not a good default - even in ideal conditions two step deserialization is noticable (bytes -> bson -> user defined struct). on the other hand i'm not sure if full separation of document::view, ::value & ::builder (c++ driver terminology) would be ergonomic enough to use with rust's borrow checker, i have only a few weeks of fighting it under my belt. so far my experience with rust's peformance was on par with c++, so i'm optimistic about the future, but i'm not sure if I can help with my current state of knowledge (for now).

jcdyer:
As a step toward this I've created #328, which removes all reference to the Bson type from the wire_protocol, without affecting any public interfaces outside that module.  A next step might involve creating a way to feed raw bson bytes to and from methods on Collections directly.  I'll give some thought as to what that should look like and put together a proposal or a PR for further discussion.

saghm:
Hello everyone! We've just released an alpha of [an official MongoDB Rust driver](https://github.com/mongodb/mongo-rust-driver); you can read about it [here](https://blog.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver)! Because of this, we won't be updating this driver anymore, and the repository will be made read-only as soon as I get finished posting these messages. Thanks for using this driver, and we hope you try out the new one!

saghm:
Apologies, the blog post link is this: https://www.mongodb.com/blog/post/announcing-the-official-mongodb-rust-driver

PR #328, opened 2019-09-27
Have the wire protocol deal with raw bytes instead of BSON documents.

jcdyer:
Keep all knowledge of structured Bson documents at the cursor level, and allow the wire protocol to deal with bson as raw Vec<u8> and &[u8] structures.  This lays the ground work for more robust solution to Issue #301.

### DIFF
--- a/src/wire_protocol/operations.rs
+++ b/src/wire_protocol/operations.rs
@@ -1,5 +1,5 @@
 //! Wire protocol operational client-server communication logic.
-use bson;
+
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use Error::{ArgumentError, ResponseError};
 use Result;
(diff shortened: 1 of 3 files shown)


src/wire_protocol/operations.rs
```diff
@@ -46,7 +30,7 @@ pub enum Message {
         /// The total number of documents being returned.
         number_returned: i32,
         /// The documents being returned.
-        documents: Vec<bson::Document>,
+        documents: Vec<u8>,
```
jcdyer: This could alternatively be a `Vec<Vec<u8>>`, where each Vec<u8> represents a single document.  The first four bytes of each document are the length of the document, so it's fairly easy to split.

src/wire_protocol/operations.rs
```diff
@@ -71,7 +55,7 @@ pub enum Message {
         /// database name and a dot separator.
         namespace: String,
         /// The documents to be inserted.
-        documents: Vec<bson::Document>,
+        documents: Vec<u8>,
```
jcdyer: This could also be a `Vec<Vec<u8>>`, and is even easier than in Query, because the present callers start with a Vec<bson::Document>, so it's just a question of serializing each to its own vec, or all to one vec.

jcdyer:
Note, this PR probably makes #326 obsolete, but 326 has less architectural impact, so I'm keeping that one open, in case it's easier to get approval on.  Both should save extra allocations per document wherever length checks are needed in the wire protocol

saghm:

PR #329, opened 2019-10-05
Fix/bind interface

allevo:
In test all assertions check the network interface too.
If your mongo binds `0.0.0.0` (for example using docker) all tests fails.
This PR fixes that.

### DIFF
--- a/tests/apm/mod.rs
+++ b/tests/apm/mod.rs
@@ -6,29 +6,8 @@ use mongodb::db::ThreadedDatabase;
 use mongodb::{Client, ClientOptions, CommandResult, ThreadedClient};
 use rand;
 
-macro_rules! assert_eq_ignoring_ip {
-    ($expected: expr, $actual: ident) => {{
-        if $actual.contains("127.0.0.1") {
-            assert_eq!($expected, &$actual);
-        } else if $actual.contains("[::1]") {
-            let expected_string = &$expected.replace("127.0.0.1", "[::1]");
-            assert_eq!(expected_string, &$actual);
-        } else {
-            unimplemented!("Unable to understand which interface is binded");
-        }
-    }};
-}
-macro_rules! assert_start_ignoring_ip {
-    ($expected: expr, $actual: ident) => {{
-        if $actual.contains("127.0.0.1") {
-            assert!($actual.starts_with($expected));
-        } else if $actual.contains("[::1]") {
-            let expected_string = &$expected.replace("127.0.0.1", "[::1]");
-            assert!($actual.starts_with(expected_string));
-        } else {
-            unimplemented!("Unable to understand which interface is binded");
-        }
-    }};
+fn coerce_ip_v6_to_v4 (line: &String) -> String {
+    line.replace("[::1]", "127.0.0.1")
 }
 
 fn timed_query(_client: Client, command_result: &CommandResult) {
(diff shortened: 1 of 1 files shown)


tests/apm/mod.rs
```diff
@@ -2,10 +2,35 @@ use std::fs::{self, File};
...
+        if $actual.contains("127.0.0.1") {
+            assert_eq!($expected, &$actual);
+        } else if $actual.contains("[::1]") {
+            let expected_string = &$expected.replace("127.0.0.1", "[::1]");
+            assert_eq!(expected_string, &$actual);
+        } else {
+            unimplemented!("Unable to understand which interface is binded");
+        }
+    }};
+}
```
jcdyer: This breaks error messaging if actual returns something that doesn't include the IP address.  It also checks for a string in actual, and then modifies expected, which could have odd results.  I'd suggest logic more like 

```rust
let actual_string = if $actual.contains("[::1]") {
    $actual.replace("[::1]", "127.0.0.1")
} else {
    $actual
};
assert_eq!($expected, actual_string);
```

tests/apm/mod.rs
```diff
@@ -2,10 +2,35 @@ use std::fs::{self, File};
 use std::io::{BufRead, BufReader};
 
 use bson::Bson;
-use mongodb::{Client, ClientOptions, CommandResult, ThreadedClient};
 use mongodb::db::ThreadedDatabase;
+use mongodb::{Client, ClientOptions, CommandResult, ThreadedClient};
 use rand;
 
+macro_rules! assert_eq_ignoring_ip {
```
jcdyer: `ignoring_ip` isn't quite an accurate name.  It's agnostic to ipv4 or ipv6, but it still expects a localhost IP address.

Maybe `assert_eq_ipv4_or_ipv6`.

tests/apm/mod.rs
```diff
@@ -107,44 +131,45 @@ fn logging() {
...
 
     // Drop collection started
     line.clear();
     read_first_non_monitor_line(&mut file, &mut line);
-    assert_eq!(
+    assert_eq_ignoring_ip!(
         "COMMAND.drop_collection 127.0.0.1:27017 STARTED: { drop: \"logging\" }\n",
-        &line
+        line
     );
```
jcdyer: An alternative implementation that might be simpler would be to introduce a `coerce_ipv4` function.

```rust
fn coerce_ipv4(val: &str) -> String {
    val.replace("[::1"], "127.0.0.1")
}
```

And then you can use the built-in assert macros:

assert_eq!(
    "COMMAND.drop_collection 127.0.0.1:27017 STARTED: { drop: \"logging\" }\n",
    coerce_ipv4(&line)   
);

jcdyer (commented):
Some suggestions.  Also note that despite the description, this won't fix IP addresses bound to 0.0.0.0, but it does fix IP addresses bound to `[::1]`.  To fix more broadly, Perhaps introduce an `ignore_ip` filter function that replaces any instance of an IPv4 or IPv6 address with a sentinel string like `<IP_ADDRESS>`, and then include that sentinel value in the expected string literals.

tests/apm/mod.rs
allevo: Nice! I like it!

allevo:
Use `coerce_ip_v6_to_v4` instead of custom assertion macros

tests/apm/mod.rs
```diff
@@ -179,53 +156,52 @@ fn logging() {
...
     assert!(line.ends_with(" ns)\n"));
 
     // Third insert started
     line.clear();
     read_first_non_monitor_line(&mut file, &mut line);
-    assert_start_ignoring_ip!(
+    assert!(
         "COMMAND.insert_one 127.0.0.1:27017 STARTED: { insert: \"logging\", documents: [{ _id: 3 }] }\n",
-        line
+        coerce_ip_v6_to_v4(&line)
```
jcdyer: This one needs to be `assert!(coerce(line).starts_with())`

allevo: Fixes

jcdyer:
Looks good to me.

@saghm / @kyeah I nominate this one to merge.  It makes testing a little more flexible, and doesn't affect non-test code.

saghm:

PR #330, opened 2019-10-05
Remove try!

allevo:
Use `?` instead of `!`.
Remove all `try!`s.

clippy warns about `try!` usage too.


jcdyer:
This looks like a straight refactor to me, and works on my projects. 

@saghm: Recommend this for merging!

saghm:

